PR #1, opened 2020-12-10
Add database setup

alex-demetriou:
### *Adding initial database setup*

- Add migration for one table, `rate_schedule_item`, using EF Core code-first approach
- Fix Makefile to include steps for updating the test-database after a new migration is created
- Remove some vestigial `lbh-base-api` stuff
- Include an initial Terraform config, not yet applied.

#### _Checklist_

- [x] Checked all code for possible refactoring
- [x] Code pipeline builds correctly

-


database/Dockerfile
```diff
@@ -1,3 +1 @@
 FROM postgres:12
-
-COPY ./database/schema.sql ./docker-entrypoint-initdb.d
```
pixeltrix, line 3: Do we need the `database/Dockerfile` file for `test-database`? We could use `image: postgres:12` like the `dev-database` entry in `docker-compose.yml`

pixeltrix (approved):

PR #7, opened 2021-01-04
Peoperties api

kylealexanderhackney:
## Link to JIRA ticket

Add a summary of the problem and any additional context that helps explain what the issue we're trying to solve is.

Add descriptions of what changes were made to address the problem and provide a solution.

#### _Checklist_


RepairsApi/V1/Gateways/ApiGateway.cs
```diff
@@ -17,12 +18,16 @@ public ApiGateway(IHttpClientFactory clientFactory)
         }
 
         [SuppressMessage("Reliability", "CA2000:Dispose objects before losing scope", Justification = "Disposal of httpclient from the factory is handled by the factory")]
```
LBHBomotoso, line 20: Is it better to suppress these in the editor config?

RepairsApi/V1/Resources.resx
```diff
@@ -0,0 +1,135 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
LBHBomotoso, line 1: Are these error messages that get returned on an error rather than hardcoded in the source?

LBHBomotoso (approved):
Looks good

RepairsApi/V1/Gateways/AlertsGateway.cs
```diff
@@ -11,40 +13,52 @@ namespace RepairsApi.V1.Gateways
...
+            {
+                _logger.LogError($"Call to {url} failed with {response.Status}");
+                throw new PlatformApiException(response.Status);
+            }
+
             return response.Content.ToDomain();
         }
 
-        public async Task<PersonAlertList> GetPersonAlertsAsync(string tenancyReference)
+        public async Task<PersonAlertList> GetPersonAlertsAsync(string? tenancyReference)
```
LBHBomotoso, line 47: Is there a difference in declaring this reference as nullable?

kylealexanderhackney: Ive enabled nullable reference types in the domain areas so this now accurately reflects what the variable can be

RepairsApi/V1/Resources.resx
kylealexanderhackney: Yes Ive added some errors to the responses and the code styles dictate const string should be in a resx file for legiblity

RepairsApi/V1/Gateways/ApiGateway.cs
kylealexanderhackney: I would say that we should keep the error turned on. The http client and its factory or something of a special case when managing disposables. the general rule "Dispose objects before losing scope" is correct this is just an exception to that rule

RepairsApi/V1/Gateways/AlertsGateway.cs
LBHBomotoso: thanks

RepairsApi/V1/Resources.resx
LBHBomotoso: cool

RepairsApi/V1/Gateways/ApiGateway.cs
LBHBomotoso: Ok, I did read something about it. https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests.
Thanks

PR #8, opened 2021-01-05
Added named client that use aithorisation header

kylealexanderhackney:
## Link to JIRA ticket

Add a summary of the problem and any additional context that helps explain what the issue we're trying to solve is.

Add descriptions of what changes were made to address the problem and provide a solution.

#### _Checklist_

- [x] Added end-to-end (i.e. integration) tests where necessary e.g to test complete functionality of a newly added feature
- [x] Added tests to cover all new production code
- [x] Added comments to the README or updated relevant documentation _(add link to documentation)_, where necessary.
- [x] Checked all code for possible refactoring
- [x] Code pipeline builds correctly

### DIFF
--- a/RepairsApi/Startup.cs
+++ b/RepairsApi/Startup.cs
@@ -20,6 +20,7 @@ using RepairsApi.V1.Gateways;
 using RepairsApi.V1.UseCase.Interfaces;
 using RepairsApi.V1.UseCase;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using System.Net.Http.Headers;
 
 namespace RepairsApi
 {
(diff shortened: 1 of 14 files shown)


RepairsApi/Startup.cs
```diff
@@ -107,7 +108,7 @@ public void ConfigureServices(IServiceCollection services)
             });
             ConfigureDbContext(services);
 
-            services.AddHttpClient();
+            AddHttpClients(services);
             services.Configure<GatewayOptions>(Configuration.GetSection(nameof(GatewayOptions)));
```
LBHBomotoso, line 112: For serverless (AWS) you will likely need to update this to 
Environment.GetEnvironmentVariable. I'm not sure if Configuration.GetSection will pick up the value from parameter store otherwise.

LBHBomotoso: Not sure if this is out of date https://aws.amazon.com/blogs/developer/net-core-configuration-provider-for-aws-systems-manager/

LBHBomotoso (commented):
Will config.getsection pick up the parameter store values?

RepairsApi/Startup.cs
kylealexanderhackney: it should pick it up from the parameter store as it puts it into environmental variables. ASp.net will bind to environmental variable assuming they are named correctly with a double underscore indicating nesting

PR #33, opened 2021-01-19
Jobstatusupdate

alex-demetriou:
Add /api/v1/jobStatusUpdate

### DIFF
--- a/RepairsApi.Tests/V2/Controllers/RepairsControllerTests.cs
+++ b/RepairsApi.Tests/V2/Controllers/RepairsControllerTests.cs
@@ -7,15 +7,15 @@ using Microsoft.AspNetCore.Mvc;
 using Moq;
 using NUnit.Framework;
 using RepairsApi.V2.Controllers;
-using RepairsApi.V2.Infrastructure;
 using RepairsApi.V2.UseCase.Interfaces;
 using System.Threading.Tasks;
-using RepairsApi.Tests.Helpers;
 using RepairsApi.V2.Factories;
 using RepairsApi.V2.Boundary.Response;
 using RepairsApi.V2.Generated;
-using WorkOrderComplete = RepairsApi.V2.Generated.WorkOrderComplete;
+using RepairsApi.V2.Infrastructure;
 using RepairsApi.Tests.Helpers.StubGeneration;
+using WorkOrderComplete = RepairsApi.V2.Generated.WorkOrderComplete;
+using JobStatusUpdate = RepairsApi.V2.Generated.JobStatusUpdate;
 
 namespace RepairsApi.Tests.V2.Controllers
 {
@@ -26,6 +26,7 @@ namespace RepairsApi.Tests.V2.Controllers
         private Mock<IListWorkOrdersUseCase> _listWorkOrdersUseCase;
         private Generator<WorkOrder> _generator;
         private Mock<ICompleteWorkOrderUseCase> _completeWorkOrderUseCase;
+        private Mock<IUpdateJobStatusUseCase> _updateJobStatusUseCase;
 
         [SetUp]
         public void SetUp()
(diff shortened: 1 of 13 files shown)


RepairsApi.Tests/V2/Controllers/RepairsControllerTests.cs
```diff
@@ -107,6 +110,21 @@ public async Task ReturnsBadRequestWhenCantCompleteWorkOrder()
...
+        public async Task ReturnsAppropriatelyForJobStatusUpdateRequest(bool useCaseSucceeds)
+        {
+            _updateJobStatusUseCase
+                .Setup(uc => uc.Execute(It.IsAny<JobStatusUpdate>()))
+                .ReturnsAsync(useCaseSucceeds);
+
+            var response = await _classUnderTest.JobStatusUpdate(
+                new JobStatusUpdate { RelatedWorkOrderReference = new Reference { ID = "42" } });
+
+            if (useCaseSucceeds) response.Should().BeOfType<OkResult>();
```
kylealexanderhackney, line 124: Typically you want 2 seperate tests for this. 
Makes for easier reading if theres consistency across the board that all asserts in statement have to pass rather than having conditional asserts.
Also if there are 2 seperate tests its immediately obvious whats wrong from the failing test rather than having a test that can fail for multiple reasons

RepairsApi.Tests/V2/UseCase/UpdateJobStatusUseCaseTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        [Test]
+        public async Task CanUpdateJobStatus()
+        {
+            _jobStatusUpdateGateway.Setup(gateway => gateway.CreateJobStatusUpdate(It.Is<JobStatusUpdate>(jsu => jsu.RelatedWorkOrder.Id == 42)))
+                .ReturnsAsync(1);
+
+            var workOrder = _fixture.Create<WorkOrder>();
+            workOrder.Id = 42;
```
kylealexanderhackney, line 41: can we put 42 into a variable to so we can label its use

RepairsApi/V2/Controllers/RepairsController.cs
```diff
@@ -63,6 +66,22 @@ public async Task<IActionResult> WorkOrderComplete([FromBody] WorkOrderComplete
...
         }
+
+        [HttpPost]
+        [Route("/api/v2/jobStatusUpdate")]
+        public async Task<IActionResult> JobStatusUpdate([FromBody] JobStatusUpdate request)
+        {
+            try
+            {
+                var result = await _updateJobStatusUseCase.Execute(request);
+                return result ? (IActionResult) Ok() : BadRequest();
```
kylealexanderhackney, line 77: should we pass a message down here in the bad request so that callers understand why there call failed?

RepairsApi/V2/Infrastructure/RepairsContext.cs
```diff
@@ -16,6 +16,7 @@ DataImporter dataImporter
         }
 
         public DbSet<WorkOrder> WorkOrders { get; set; }
+        public DbSet<WorkElement> WorkElements { get; set; }
```
kylealexanderhackney, line 19: Is this getting used?

RepairsApi/V2/UseCase/UpdateJobStatusUseCase.cs
```diff
@@ -0,0 +1,55 @@
...
+            var workOrderId = int.Parse(request.RelatedWorkOrderReference.ID);
+
+            var workOrder = _repairsGateway.GetWorkOrder(workOrderId);
+            if (workOrder is null) return false;
+
+            var workElements = _repairsGateway.GetWorkElementsForWorkOrder(workOrder);
+
+            await _jobStatusUpdateGateway.CreateJobStatusUpdate(
+                new Infrastructure.JobStatusUpdate
+                {
```
kylealexanderhackney, line 36: is it worth wrapping this up into its own todb call?

RepairsApi.Tests/V2/Controllers/RepairsControllerTests.cs
alex-demetriou: Ah fair enough - I just thought it read nicely
In my test explorer it shows up as two tests as well so I thought this was fine
I don't mind splitting it out if that's preferred

RepairsApi/V2/Infrastructure/RepairsContext.cs
alex-demetriou: Was using it in an earlier draft - thanks for catching

josh-dadswell-hackney (approved):

kylealexanderhackney (approved):

PR #59, opened 2021-02-01
More specific sor code

josh-dadswell-hackney:
## Link to JIRA ticket

https://trello.com/c/f9d9MPHl/290-as-a-contractor-admin-i-need-to-be-able-to-vary-jobs

### DIFF
--- a/RepairsApi/V2/Gateways/RepairsGateway.cs
+++ b/RepairsApi/V2/Gateways/RepairsGateway.cs
@@ -51,5 +51,13 @@ namespace RepairsApi.V2.Gateways
 
             return await elements.SingleOrDefaultAsync();
         }
+
+        public async Task AddWorkElement(int id, WorkElement workElement)
+        {
+            var workOrder = await GetWorkOrder(id);
+            workOrder.WorkElements.Add(workElement);
+
+            await _repairsContext.SaveChangesAsync();
+        }
     }
 }
--- a/RepairsApi/V2/Gateways/RepairsGateway.cs
+++ b/RepairsApi/V2/Gateways/RepairsGateway.cs
@@ -44,9 +44,14 @@ namespace RepairsApi.V2.Gateways
 
         public async Task<IEnumerable<WorkElement>> GetWorkElementsForWorkOrder(WorkOrder workOrder)
         {
-            var elements =
+            return await GetWorkElementsForWorkOrder(workOrder.Id);
+        }
+
+        public async Task<IEnumerable<WorkElement>> GetWorkElementsForWorkOrder(int id)
+        {
+            IQueryable<List<WorkElement>> elements =
                 from wo in _repairsContext.WorkOrders
-                where wo.Id == workOrder.Id
+                where wo.Id == id
                 select wo.WorkElements;
 
             return await elements.SingleOrDefaultAsync();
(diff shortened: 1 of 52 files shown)


RepairsApi/V2/Gateways/RepairsGateway.cs
```diff
@@ -56,5 +56,13 @@ public async Task<IEnumerable<WorkElement>> GetWorkElementsForWorkOrder(int id)
 
             return await elements.SingleOrDefaultAsync();
         }
+
+        public async Task AddWorkElement(int id, WorkElement workElement)
+        {
+            var workOrder = await GetWorkOrder(id);
```
kylealexanderhackney, line 62: Set the id on the work element to avoid a query

RepairsApi/V2/UseCase/JobStatusUpdatesUseCases/MoreSpecificSorUseCase.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var workOrder = await _repairsGateway.GetWorkOrder(workOrderId);
+
+            if (workOrder == null)
+            {
+                throw new ResourceNotFoundException($"Unable to locate work order {workOrderId}");
+            }
+
+            var existingCodes = workOrder.WorkElements.SelectMany(we => we.RateScheduleItem);
+            var newCodes = workElement.RateScheduleItem.Where(rsi => !existingCodes.Any(ec => ec.CustomCode == rsi.CustomCode));
```
kylealexanderhackney, line 37: there is an except linq command which may resolve to a better query

RepairsApi/V2/UseCase/UpdateJobStatusUseCase.cs
```diff
@@ -31,9 +37,21 @@ public async Task<bool> Execute(JobStatusUpdate jobStatusUpdate)
...
             await _jobStatusUpdateGateway.CreateJobStatusUpdate(jobStatusUpdate.ToDb(workElements, workOrder));
 
             return true;
         }
+
+        private async Task ProcessActions(JobStatusUpdate jobStatusUpdate)
+        {
+            switch (jobStatusUpdate.TypeCode)
+            {
+                case JobStatusUpdateTypeCode._80: // More specific SOR Code
```
kylealexanderhackney, line 51: we should have a default case where we may be throw a not supported to indicate the supported codes

kylealexanderhackney (changes requested):

RepairsApi/V2/UseCase/UpdateJobStatusUseCase.cs
```diff
@@ -31,9 +37,21 @@ public async Task<bool> Execute(JobStatusUpdate jobStatusUpdate)
...
+                    await _moreSpecificSorUseCase.Execute(jobStatusUpdate);
```
kylealexanderhackney, line 52: this should probably be done via a factory class creating a the use case as a strategy. just to avoid having to inject loads of use cases into this class in the future. 
structure could be a factory that inject the service provider. given an id it then creates the correct use case and executes it.

kylealexanderhackney (changes requested):

kylealexanderhackney (approved):

PR #108, opened 2021-02-22
Update master

kylealexanderhackney:
Updating master with security checks

This work requires a new aws parameter that is the json for what google groups are for which contractor


LBHMKeyworth (approved):
Looks good to me.

PR #124, opened 2021-02-24
Added ids and fixed security to work order

kylealexanderhackney:
This PR moves the update of sor codes auhorisation check later so its only executed for that particular type of update.
It also adds a UID for rate schedule items so that duplicate items with the sor code on variation.

This needs to be lined up with f/e work to receive the new ids and pass them on when performing an update of the sor codes


LBHMKeyworth (approved):
Looks good

PR #130, opened 2021-03-01
Added appointment info to work order response

kylealexanderhackney:
Adds appointment information to the work order response

### DIFF
--- a/RepairsApi/V2/DateConstants.cs
+++ b/RepairsApi/V2/DateConstants.cs
@@ -1,7 +1,15 @@
+using System;
+using System.Globalization;
+
 namespace RepairsApi.V2
 {
     public static class DateConstants
     {
         public const string DATEFORMAT = "yyyy-MM-dd";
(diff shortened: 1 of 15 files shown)


RepairsApi/V2/DateConstants.cs
```diff
@@ -1,7 +1,15 @@
...
+
 namespace RepairsApi.V2
 {
     public static class DateConstants
     {
         public const string DATEFORMAT = "yyyy-MM-dd";
+
+        public static string ToISO(this DateTime date)
+        {
+            return date.ToString("o", CultureInfo.InvariantCulture);
```
LBHMKeyworth, line 12: I'll be intrigued to see the result from this because at the back of my mind I seem to remember that Windows and Linux treated dates differently when formatting them for daylight savings but I might be imagining it...

LBHMKeyworth (approved):
Looks good.

PR #131, opened 2021-03-01
Add JSU support for holding and resuming work orders

kylealexanderhackney:
job status update now supported resolving job incomplete and a custom generic resume job update

### DIFF
--- a/RepairsApi/V2/Infrastructure/Extensions/WorkOrderExtensions.cs
+++ b/RepairsApi/V2/Infrastructure/Extensions/WorkOrderExtensions.cs
@@ -6,13 +6,14 @@ namespace RepairsApi.V2.Infrastructure.Extensions
     {
         public static string GetStatus(this WorkOrder workOrder)
         {
-            switch (workOrder.StatusCode)
+            return workOrder.StatusCode switch
             {
-                case WorkStatusCode.Open: return WorkOrderStatus.InProgress;
-                case WorkStatusCode.Complete: return WorkOrderStatus.Complete;
-                case WorkStatusCode.Canceled: return WorkOrderStatus.Cancelled;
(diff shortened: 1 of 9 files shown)


RepairsApi/V2/Infrastructure/Extensions/WorkOrderExtensions.cs
```diff
@@ -6,13 +6,14 @@ public static class WorkOrderExtensions
...
-                case WorkStatusCode.Complete: return WorkOrderStatus.Complete;
-                case WorkStatusCode.Canceled: return WorkOrderStatus.Cancelled;
-                default: return WorkOrderStatus.Unknown;
-            }
+                WorkStatusCode.Open => WorkOrderStatus.InProgress,
+                WorkStatusCode.Complete => WorkOrderStatus.Complete,
+                WorkStatusCode.Canceled => WorkOrderStatus.Cancelled,
+                WorkStatusCode.Hold => WorkOrderStatus.Hold,
+                _ => WorkOrderStatus.Unknown,
+            };
```
LBHMKeyworth, line 16: I'd not see that way of doing a switch before...nice!

LBHMKeyworth (approved):
Looks good.

RepairsApi/V2/Infrastructure/Extensions/WorkOrderExtensions.cs
kylealexanderhackney: Yh it was an IDE suggestion but I prefer how it looks. switches seem to be becoming friendlier to use and even support if style check with the when keyword
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/switch#the-case-statement-and-the-when-clause

PR #146, opened 2021-03-09
Create variations and users migration

LBHBomotoso:
## Link to JIRA ticket

https://trello.com/c/8ofepMTh/21-as-a-contractor-i-am-informed-that-a-work-order-is-over-my-variation-limit-and-will-be-submitted-for-approval

Please refer to ticket

Added variation group and user entity for ef migration

#### _Checklist_


kylealexanderhackney (changes requested):
Is there a reason this cant be managed with the security_groups table.
adding a row to that table with only the group_name and vary_limit columns set give users with that group that mapped vary limit (this is already integrated into variations). Also i would be careful storing users in our system. in a microservice architecture users pretty much always have their own service or its just managed by the IDP (in our case google)

PR #147, opened 2021-03-11
Hub user endpoint

LBHBomotoso:
## Link to JIRA ticket

https://trello.com/c/8ofepMTh/21-as-a-contractor-i-am-informed-that-a-work-order-is-over-my-variation-limit-and-will-be-submitted-for-approval

See ticket

created /api/v2/hub-user end point and tests

#### _Checklist_

### DIFF
--- a/RepairsApi.Tests/HttpClientExtensions.cs
+++ b/RepairsApi.Tests/HttpClientExtensions.cs
@@ -19,6 +19,7 @@ namespace RepairsApi.Tests
                 .AddClaim("email", user.Email)
                 .AddClaim("name", user.Name)
                 .AddClaim("groups", user.Groups)
+                .AddClaim("sub", user.Sub)
                 .Encode();
             client.DefaultRequestHeaders.Remove("X-Hackney-User");
             client.DefaultRequestHeaders.Add("X-Hackney-User", jwt);
@@ -28,6 +29,7 @@ namespace RepairsApi.Tests
         {
             client.SetUser(new RepairsApi.V2.Domain.User
             {
+                Sub = TestUserInformation.SUB,
                 Name = TestUserInformation.NAME,
                 Email = TestUserInformation.EMAIL,
                 Groups = new List<string>()
@@ -43,6 +45,7 @@ namespace RepairsApi.Tests
         {
             client.SetUser(new RepairsApi.V2.Domain.User
             {
+                Sub = TestUserInformation.SUB,
                 Name = TestUserInformation.NAME,
                 Email = TestUserInformation.EMAIL,
                 Groups = new List<string>()
--- a/RepairsApi.Tests/HttpClientExtensions.cs
+++ b/RepairsApi.Tests/HttpClientExtensions.cs
@@ -41,6 +41,22 @@ namespace RepairsApi.Tests
             });
         }
 
+        public static void SetAgent(this HttpClient client, string raiseLimit, string varyLimit)
+        {
+            client.SetUser(new RepairsApi.V2.Domain.User
+            {
+                Sub = TestUserInformation.SUB,
+                Name = TestUserInformation.NAME,
+                Email = TestUserInformation.EMAIL,
+                Groups = new List<string>()
+                {
+                    "agent",
+                    raiseLimit,
+                    varyLimit
+                }
+            });
+        }
+
         public static void SetGroup(this HttpClient client, string group)
         {
             client.SetUser(new RepairsApi.V2.Domain.User
--- a/RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
+++ b/RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
@@ -18,5 +18,15 @@ namespace RepairsApi.V2.Authorisation
         {
             return claimsPrincipal.FindAll(CustomClaimTypes.CONTRACTOR).Select(c => c.Value).ToList();
         }
+
+        public static string VaryLimit(this ClaimsPrincipal claimsPrincipal)
+        {
+            return claimsPrincipal.FindFirst(CustomClaimTypes.VARYLIMIT).Value;
+        }
+
+        public static string RaiseLimit(this ClaimsPrincipal claimsPrincipal)
+        {
+            return claimsPrincipal.FindFirst(CustomClaimTypes.RAISELIMIT).Value;
+        }
     }
 }
--- a/RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
+++ b/RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
@@ -28,5 +28,10 @@ namespace RepairsApi.V2.Authorisation
         {
             return claimsPrincipal.FindFirst(CustomClaimTypes.RAISELIMIT).Value;
         }
+
+        public static string Sub(this ClaimsPrincipal claimsPrincipal)
+        {
+            return claimsPrincipal.FindFirst(ClaimTypes.PrimarySid).Value;
+        }
     }
 }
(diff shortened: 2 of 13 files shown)


RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
```diff
@@ -18,5 +18,20 @@ public static List<string> Groups(this ClaimsPrincipal claimsPrincipal)
         {
             return claimsPrincipal.FindAll(CustomClaimTypes.CONTRACTOR).Select(c => c.Value).ToList();
         }
+
+        public static string VaryLimit(this ClaimsPrincipal claimsPrincipal)
+        {
```
kylealexanderhackney, line 23: these look like they could all share a private Utility method called get claim or something. 
you could then potentially clean up using expression bodies members
so youd get something like 
`public static string RaiseLimit(this ClaimsPrincipal claimsPrincipal) => GetClaim(claimsPrincipal, CustomClaimTypes.VARYLIMIT)`

RepairsApi/V2/Boundary/Response/HubUserModel.cs
```diff
@@ -0,0 +1,12 @@
+namespace RepairsApi.V2.Boundary.Response
+{
+    public class HubUserModel
```
kylealexanderhackney, line 3: Hub by itself doesnt help much to someone looking at the code for the first time. i think RepairsHubUserModel would be clearer if we want to be explicit but UserModel makes sense aswell

RepairsApi/V2/Controllers/HubUserController.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Gets repairs hub user
+        /// </summary>
+        /// <returns></returns>
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(HubUserModel), 200)]
+        [HttpGet]
+        [Authorize(Roles = UserGroups.AGENT + "," + UserGroups.CONTRACTOR)]
+        public async Task<ActionResult> GetHubUser()
+        {
+            return Ok(await Task.FromResult(_currentUserService.GetHubUser()));
```
kylealexanderhackney, line 33: You should be able to avoid the current user service here and just get the data straight from HTTPContext.User

RepairsApi.Tests/HttpClientExtensions.cs
```diff
@@ -19,6 +19,7 @@ public static void SetUser(this HttpClient client, User user)
                 .AddClaim("email", user.Email)
                 .AddClaim("name", user.Name)
                 .AddClaim("groups", user.Groups)
+                .AddClaim("sub", user.Sub)
```
kylealexanderhackney, line 22: Is sub a field the front end definitely needs? ideally endpoints that ship user info should only be sending down the minimum needed by the frontend to avoid. Its not a problem to have it but keeping the info on these endpoint to a minimum is a good rule of thumb

kylealexanderhackney (changes requested):

RepairsApi/V2/Services/CurrentUserService.cs
```diff
@@ -60,12 +61,31 @@ public bool TryGetContractor(out string contractor)
             return !string.IsNullOrWhiteSpace(contractor);
         }
 
+        public HubUserModel GetHubUser()
+        {
+            var hubUser = new HubUserModel();
+
+            hubUser.Sub = _user.FindFirstValue(ClaimTypes.PrimarySid);
```
kylealexanderhackney, line 68: you can use the methods you added to ClaimsPrincipalExtension to simplify this.

RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
LBHBomotoso: Change the whole class or just the 2 methods? These are only used in the tests if I recall. I prefer the explicit statements for readability in C#

RepairsApi/V2/Boundary/Response/HubUserModel.cs
LBHBomotoso: ok

RepairsApi/V2/Controllers/HubUserController.cs
LBHBomotoso: We need the translation from HttpContext user to hub user, thats better done in a service layer right?

RepairsApi.Tests/HttpClientExtensions.cs
LBHBomotoso: Sub was requested

RepairsApi/V2/Services/CurrentUserService.cs
LBHBomotoso: This is a .net function. It just returns the value being requested.

RepairsApi/V2/Controllers/HubUserController.cs
kylealexanderhackney: potentially IMO the service layer is for storing business logic and a data model transformation doesnt really qualify in my opinion but its fine either way. the general counter point is just keeping the code as simple to new people and avoiding adding extra layers needlessly.

LBHBomotoso: It seemed to be handling all the user functions and creates a user object which is mapped via the decrypted JWT.

RepairsApi/V2/Authorisation/ClaimsPrincipalExtension.cs
kylealexanderhackney: It would just be adding in a small method to shorten calls
`private static string GetClaim(ClaimsPrincipal claimsPrincipal, string claimName) => 
 claimsPrincipal.FindFirst(claimName).Value;`

Its just a small QOL change to anticipate more fields getting added and this class potentially looking a bit messy. if all the methods are expression bodied then the class becomes nicer to read and understand . in my opinion anyway

RepairsApi/V2/Controllers/HubUserController.cs
kylealexanderhackney: ATM its a stateful scoped service whose responsibility is serving the current user. the transformation method is stateless and adds dependencies to new models that dont really need to be there. it could be written as an extension method which i guess solves both issues. I guess the hackney standard says it should go in a factory class as an extension method

RepairsApi/V2/Controllers/HubUserController.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        /// <summary>
+        /// Gets repairs hub user
+        /// </summary>
+        /// <returns></returns>
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(HubUserModel), 200)]
+        [HttpGet]
+        [Authorize(Roles = UserGroups.AGENT + "," + UserGroups.CONTRACTOR)]
```
kylealexanderhackney, line 30: Authorize doesnt need to be here any role added to the system in the future can also call this method. anonymous access is handled up front by aws so its not an issue

RepairsApi/V2/Services/CurrentUserService.cs
kylealexanderhackney: yh i just mean the purpose of the claims principal extensions is to simplify this call and also make it easier to understand what claims the user object is expected to have.
if there is a _user.Sub() it helps someone maintaining the system understand there will be a sub loaded without having to check around and look for what claimtypes are used

RepairsApi/V2/Controllers/HubUserController.cs
LBHBomotoso: ok

LBHBomotoso: removed

RepairsApi/V2/Controllers/HubUserController.cs
LBHBomotoso: amended

kylealexanderhackney (approved):

PR #150, opened 2021-03-15
Allow filtering by work order status

LBHMKeyworth:
## Link to Trello ticket

https://trello.com/c/RaKeUK2Q/26-as-a-contract-manager-i-can-log-in-and-see-a-list-of-orders-so-that-i-know-which-ones-i-need-to-action

## Describe this PR

Work orders need to be filtered by status

When listing work orders we need to be able filter by status

Added another option to WorkOrderSearchParameters on GET /api/v2/repairs/.
Added extra param to the Gateway search.
Added test for gw.


alex-demetriou (approved):
👍

PR #182, opened 2021-04-13
Configure AWS XRay

alex-demetriou:
## [Trello: Add AWS X-Ray to API](https://trello.com/c/v4m5P1rd)

Configures AWS X-Ray for the API

### DIFF
--- a/RepairsApi/RepairsApi.csproj
+++ b/RepairsApi/RepairsApi.csproj
@@ -16,6 +16,10 @@
 
   <ItemGroup>
     <PackageReference Include="Amazon.Lambda.AspNetCoreServer" Version="5.1.1" />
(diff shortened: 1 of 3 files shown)


kylealexanderhackney (approved):
Are we aware of the performance impact of this?. I assume its not much given its wide usage but worth understanding nonetheless

alex-demetriou:
> Are we aware of the performance impact of this?. I assume its not much given its wide usage but worth understanding nonetheless

Good question, yeah - I found [this conversation](https://forums.aws.amazon.com/thread.jspa?threadID=273718) with an AWS rep  
> [...] every effort has been made to minimize the overhead. That being said, precise statements about performance of the X-Ray SDKs are difficult to make.

Also seems like people aren't having issues with it, going by the lack of discussion around it

PR #188, opened 2021-04-15
Filtering

kylealexanderhackney:
Refactored filtering and added filtering by multiple statuses


RepairsApi.Tests/V2/E2ETests/RepairApiTests.cs
```diff
@@ -99,6 +99,28 @@ public async Task GetListOfWorkOrders()
             response.Should().Contain(wo => wo.Reference == workOrderId);
         }
 
+        [Test]
+        public async Task GetFilteredListOfWorkOrders()
```
josh-dadswell-hackney, line 103: Needs to test both are returned

josh-dadswell-hackney (approved):

PR #205, opened 2021-04-22
Use "Variation Pending Approval"

alex-demetriou:
## [Trello: As a contract manager, I can see if a request is variation pending approval, so that I know which repairs to approve](https://trello.com/c/uWW6pSKg)

 - Changes *Pending Approval* to *Variation Pending Approval* in `WorkOrderStatus`.
 - Changes *Pending Authorisation* to *Variation Pending Authorisation* in `WorkOrderReason`.

### DIFF
--- a/RepairsApi/V2/Boundary/Response/WorkOrderListItem.cs
+++ b/RepairsApi/V2/Boundary/Response/WorkOrderListItem.cs
@@ -8,7 +8,7 @@ namespace RepairsApi.V2.Boundary.Response
         public const string Complete = "Work Complete";
         public const string Cancelled = "Work Cancelled";
         public const string Hold = "On Hold";
-        public const string PendApp = "Pending Approval";
+        public const string PendApp = "Variation Pending Approval";
         public const string PendMaterial = "Materials Needed";
         public const string VariationApproved = "Variation Approved";
         public const string VariationRejected = "Variation Rejected";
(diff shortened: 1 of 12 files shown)


LBHBomotoso (approved):

kylealexanderhackney (approved):

RepairsApi/V2/Boundary/Response/WorkOrderListItem.cs
```diff
@@ -8,7 +8,7 @@ public static class WorkOrderStatus
         public const string Complete = "Work Complete";
         public const string Cancelled = "Work Cancelled";
         public const string Hold = "On Hold";
-        public const string PendApp = "Pending Approval";
+        public const string PendApp = "Variation Pending Approval";
```
kylealexanderhackney, line 11: WE should change the variable names also in my opinion.
Potentially even the work status enum. as long as the number stays the saem there should be no problem  but the variable names should reflect the fact its specifically for variations as well

kylealexanderhackney (changes requested):

RepairsApi/V2/Boundary/Response/WorkOrderListItem.cs
alex-demetriou: Cheers - that's a good point!
I've pushed these changes now

kylealexanderhackney (approved):

PR #248, opened 2021-05-12
Company tmo table

elena-evans:
## Link to JIRA ticket

https://trello.com/c/dVNIA7Ky/118-as-a-user-i-can-see-if-this-is-a-tmo-property-so-that-repairs-are-not-raised-in-error

Added class for TMO company table and migration

Created lookup table for TMO data

#### _Checklist_

Load data into tables


RepairsApi/V2/Infrastructure/Migrations/20210512104555_TMOCompanyTable.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace RepairsApi.V2.Infrastructure.Migrations
+{
+    public partial class TMOCompanyTable : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
```
josh-dadswell-hackney, line 7: In order for EFCore to know to create the table the new type must be referenced either by another object in the DB or manually setup in the database context.

There is a simple example here: https://www.entityframeworktutorial.net/efcore/entity-framework-core-dbcontext.aspx#:~:text=The%20DbContext%20class%20is%20an,Of%20Work%20and%20Repository%20patterns.

Also when generating a new Migration be sure to remove that empty one first

josh-dadswell-hackney (changes requested):
Migration incomplete

RepairsApi/V2/Infrastructure/Migrations/20210512104555_TMOCompanyTable.cs
kylealexanderhackney: Simplest would be to add a dbset for this object

kylealexanderhackney:
Is there a reason the work required for TMO's cant be done in tandem with this and just merge the feature in complete

RepairsApi/V2/Infrastructure/Company.cs
```diff
@@ -0,0 +1,25 @@
+using System.ComponentModel.DataAnnotations;
+
+namespace RepairsApi.V2.Infrastructure
+{
+    public class Company
+    {
+        [StringLength(10)]
```
kylealexanderhackney, line 7: missing setup for a key. AFAIK the CoCode needs a [Key] attribute if its is the primary key

kylealexanderhackney (changes requested):

RepairsApi/V2/Infrastructure/Company.cs
```diff
@@ -0,0 +1,25 @@
...
+        [StringLength(50)]
+        public string Name { get; set; }
+        [StringLength(10)]
+        public string CloneOf { get; set; }
+        public char CGroup { get; set; }
+        [StringLength(200)]
+        public string CompAvail { get; set; }
+        [StringLength(200)]
+        public string CompDisplay { get; set; }
+        public bool CoEnabled { get; set; }
```
kylealexanderhackney, line 22: I believe primitives are created as non nullable by ef core unless they are marked as nullable (eg bool?)

does this line up with the data we will be migrating in?

RepairsApi/V2/Infrastructure/Migrations/20210512104555_TMOCompanyTable.cs
LBHBomotoso: Yes, forgot to add the dbset. Will be added

RepairsApi/V2/Infrastructure/Company.cs
LBHBomotoso: we can make it unique

RepairsApi/V2/Infrastructure/Company.cs
LBHBomotoso: yes it does, thanks

RepairsApi/V2/Infrastructure/Company.cs
kylealexanderhackney: It will need to be unique if we are going ot mark it as the primary key

RepairsApi/V2/Infrastructure/Migrations/20210514143056_TMOCompanyTable.cs
```diff
@@ -0,0 +1,17 @@
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace RepairsApi.V2.Infrastructure.Migrations
+{
+    public partial class TMOCompanyTable : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
```
kylealexanderhackney, line 8: a migration runs it is these methods that change the db. if they are empty no change will occur to the database. for what you are doing i would expect to see something similar to https://github.com/LBHackney-IT/repairs-api-dotnet/blob/develop/RepairsApi/V2/Infrastructure/Migrations/20201210131629_AddRateScheduleItem.cs
it seems everything is setup right the migration itself just needs to be removed and readded via the cli

kylealexanderhackney (changes requested):

kylealexanderhackney (approved):

PR #260, opened 2021-05-21
Provide endpoints for Operatives data

alex-demetriou:
## Trello
 - ####  [Add operatives tables to the db](https://trello.com/c/Rd7NUWCJ)
 - ####  [As a user I can add an operative to a work order so that they are associated with that WO](https://trello.com/c/1PPQVvM4)

## Describe this PR
 - Adds Operative endpoints
 - Modifies the Operatives table to house data from DRS

## Progress
- [x] List endpoint
- [x] Get endpoint
- [x] Delete endpoint (soft delete)
- [x] Table migrations
- [x] Register classes in `Startup.cs`
- [ ] End-to-end tests (?)


RepairsApi/V2/Boundary/Request/Operative.cs
```diff
@@ -0,0 +1,10 @@
+namespace RepairsApi.V2.Boundary.Request
+{
+    public class Operative
```
josh-dadswell-hackney, line 3: Can we rename this class to OperativeRequest, will also save the 'using alias'

RepairsApi/V2/Boundary/Response/Operative.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+
+namespace RepairsApi.V2.Boundary.Response
+{
+    public class Operative
```
josh-dadswell-hackney, line 5: Probably should rename this also operativeResponse

RepairsApi/V2/Controllers/OperativesController.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Retrieves all matching operatives given the query params
+        /// </summary>
+        /// <param name="operativeSearchParams">The search parameters object to match against</param>
+        /// <response code="200">Operatives found</response>
+        [HttpGet]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(List<Operative>), 200)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status502BadGateway)]
```
josh-dadswell-hackney, line 34: I dont think this endpoint will ever return a 502 as it doesnt use any external APIs

kylealexanderhackney, line 34: 502s are only returned if this endpoint calls another endpoint which can fail. I dont believe this work does that

RepairsApi/V2/Gateways/OperativeGateway.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public OperativeGateway(RepairsContext context)
+        {
+            _context = context;
+        }
+
+        public async Task<IEnumerable<Operative>> GetByQueryAsync(Boundary.Request.Operative searchModel)
+        {
+            var query = _context.Operatives
+                .Where(operative => searchModel.Id == null || operative.Id == searchModel.Id)
```
kylealexanderhackney, line 22: We have a framework for providing a filter framework on queries that simplifies this class and therefore test maintenence.

Im happy to jump into a call about how it works and work order list is a good reference
ListWorkOrdersUseCase line 29 is the start of how its used
and FilteringServiceCollectionExtensions is how it is configured

alex-demetriou: I've refactored to use the Filtering Service - could you have a look? :pray:

kylealexanderhackney (approved):
All looks good. Only thing i would mention is that an integration test that archives and then gets operatives to prove the archive works would be helpful.

alex-demetriou:
> All looks good. Only thing i would mention is that an integration test that archives and then gets operatives to prove the archive works would be helpful.

Cheers :pray: 
Would that be significantly different to [this](https://github.com/LBHackney-IT/repairs-api-dotnet/blob/21695983915cde05ddb3768438c9c475066b85b6/RepairsApi.Tests/V2/Gateways/OperativeGatewayTests.cs#L52-L70)?

RepairsApi/V2/Infrastructure/Operative.cs
```diff
@@ -1,11 +1,14 @@
...
 using System.ComponentModel.DataAnnotations;
+using Microsoft.EntityFrameworkCore;
 
 namespace RepairsApi.V2.Infrastructure
 {
-    public class Operative
+    public class Operative : IArchivable
     {
         [Key] public int Id { get; set; }
+        public string PayrollNumber { get; set; }
```
josh-dadswell-hackney, line 10: Is there any reason the payroll number is not the primary key?

PR #264, opened 2021-05-25
Add tmo data to property object response

elena-evans:
## Link to JIRA ticket

https://trello.com/c/v5YU3xU8/99-check-all-tmo-properties-are-listed-correctly

added TMO to property model

added TMO to property model and tests for this

#### _Checklist_


.editorconfig
```diff
@@ -152,3 +152,6 @@ dotnet_diagnostic.CA1707.severity = none
 
 # CA1308: Normalize strings to uppercase
 dotnet_diagnostic.CA1308.severity = none
+
+# CS8602: Dereference of a possibly null reference
+dotnet_diagnostic.CS8602.severity = none
```
kylealexanderhackney, line 157: We shouldnt supress this its a useful feature for understanding when a variable will be null

RepairsApi.Tests/V2/Gateways/PropertyGatewayTests.cs
```diff
@@ -6,6 +6,7 @@
 using RepairsApi.V2.Exceptions;
```
kylealexanderhackney, line 6: Can we put some tests in here to check a "happy path". IE it does match a TMO and we see that TMO in the object

kylealexanderhackney (changes requested):

.editorconfig
elena-evans: We tried avoiding triggering this rule in the code by changing it to
`string? tmo = _repairsContext.Company.Where(c => c.CoCode.Equals(response.Content.CompAvail))                .SingleOrDefault()?.Name;`
but it still triggered it. Do you have an idea for how else we could avoid that?

kylealexanderhackney: response.Content!.CompAvail will do it. its complaining because response.Content is not guaranteed to not be null. There is a success check above this code so we can use the exclamation mark to tell the compile to ignore the rule in this statement

elena-evans: Thanks Kyle, I changed it to that and it was happy!

RepairsApi.Tests/V2/Gateways/PropertyGatewayTests.cs
elena-evans: We've added this in now

RepairsApi.Tests/V2/Gateways/PropertyGatewayTests.cs
```diff
@@ -40,6 +41,28 @@ public async Task SendsRequestById()
...
+        [TestCase("002", "TMO Name2")]
+        [TestCase("003", "TMO Name3")]
+        public async Task SendsRequestByIdAndGetsTMOName(string comp, string name)
+        {
+            // Arrange
+            var stubData = BuildResponse(StubPropertyApiResponse().Generate());
+            stubData.Content.CompAvail = comp;
+            _apiGatewayMock.Setup(gw => gw.ExecuteRequest<PropertyApiResponse>(It.IsAny<string>(), It.IsAny<Uri>())).ReturnsAsync(stubData);
+
+
```
josh-dadswell-hackney, line 58: whitespace

.editorconfig
```diff
@@ -152,3 +152,4 @@ dotnet_diagnostic.CA1707.severity = none
 
 # CA1308: Normalize strings to uppercase
 dotnet_diagnostic.CA1308.severity = none
+
```
josh-dadswell-hackney, line 155: should revert these before commiting

RepairsApi.Tests/TestDataSeeder.cs
```diff
@@ -69,6 +69,7 @@ public static void SeedData(this RepairsContext ctx)
                     SeedSorMap(ctx);
 
                     SeedSecurityGroups(ctx);
+
```
josh-dadswell-hackney, line 72: needless whitespace

RepairsApi.Tests/V2/Gateways/PropertyGatewayTests.cs
```diff
@@ -40,6 +41,28 @@ public async Task SendsRequestById()
 
             // Assert
             result.Address.ShortAddress.Should().Be(stubData.Content.Address1);
+            result.TmoName.Should().Be(null);
+        }
+
+        [Test]
+        [TestCase("001", "TMO Name1")]
```
josh-dadswell-hackney, line 48: Usually reserve multiple test cases for testing different parts of the code, this looks like the extra tests arent adding any value so are not worth the execution time

RepairsApi.Tests/V2/Gateways/PropertyGatewayTests.cs
```diff
@@ -73,6 +96,23 @@ public async Task ThrowsFor404()
             await actFunction.Should().ThrowAsync<ResourceNotFoundException>();
         }
 
+        private static void SeedCompany()
+        {
+            InMemoryDb.Instance.Company.AddRange(new List<Company>
+            {
+                new Company {CoCode = "001", CompAvail = "001", Description = "TMO Address1", Name = "TMO Name1" },
```
josh-dadswell-hackney, line 103: should hold the test data somewhere central (class variable) so that you dont have magic strings here and in tests e.g. "001" etc.

RepairsApi.Tests/V2/UseCase/GetPropertyUseCaseTests.cs
```diff
@@ -34,6 +34,7 @@ public async Task ReturnsProperty()
         {
             // Arrange
             var expectedProperty = StubProperties().Generate();
+            var tmo = expectedProperty.TmoName;
             var expectedLocationAlerts = StubPropertyAlertList(expectedProperty.PropertyReference, 5);
             var expectedPersonAlerts = new PersonAlertList() { Alerts = StubAlerts().Generate(5) };
```
josh-dadswell-hackney, line 39: magic strings should be in vars

RepairsApi/V2/Gateways/Models/PropertyApiResponse.cs
```diff
@@ -115,7 +117,6 @@ public class PropertyApiResponse
         public bool? SecToileta { get; set; }
         public string SecToiletb { get; set; }
         public string CookingFuel { get; set; }
-        public string CompAvail { get; set; }
         public string CompDisplay { get; set; }
         public short NoSingleBeds { get; set; }
         public short NoDoubleBeds { get; set; }
```
josh-dadswell-hackney, line 122: I know its not all from this PR but why is there this commented out code? Should not leave stuff like this in the code base as its confusing

josh-dadswell-hackney (changes requested):

kylealexanderhackney (approved):

josh-dadswell-hackney (approved):

PR #267, opened 2021-05-26
Add ResourcePolicy for DRS Receiver SOAP endpoint

alex-demetriou:
Adds Resource Policy for the DRS Receiver SOAP endpoint
> DenyAll -> Allow DRS provided IPs


teodor-hristoforov-hackney:
@Asura5 @pixeltrix Is this still something we want in main branch or should we remove this PR?

PR #269, opened 2021-06-02
Add an operative manager role

pixeltrix:


RepairsApi/V2/Controllers/OperativesController.cs
```diff
@@ -56,8 +56,8 @@ public async Task<IActionResult> GetOperative([FromRoute][Required] string opera
         [Produces("application/json")]
         [ProducesResponseType(typeof(List<OperativeRequest>), 200)]
         [ProducesDefaultResponseType]
-        [Authorize(Roles = UserGroups.Agent + "," + UserGroups.ContractManager + "," + UserGroups.AuthorisationManager)]
-        public async Task<IActionResult> ListOperatives(OperativeRequest operativeRequest)
+        [Authorize(Roles = UserGroups.OperativeManager)]
```
Asura5, line 59: Just confirming that this endpoint is not used by existing users, and if it is, that these users have been given this role?

pixeltrix: No, it's still only deployed to the development environment - roles were copy/pasted from another endpoint

RepairsApi/Dockerfile
```diff
@@ -3,8 +3,10 @@ FROM mcr.microsoft.com/dotnet/core/sdk:3.1
 WORKDIR /app
 
 # Copy csproj and restore as distinct layers
-COPY *.csproj ./
-RUN dotnet restore
+COPY ./RepairsApi.sln ./
```
kylealexanderhackney, line 6: Are these changes necessary?

pixeltrix: To be able to run `make serve` using Docker on a Mac, yes. Otherwise you have to run it through Visual Studio.

alex-demetriou (approved):

PR #274, opened 2021-06-04
Small: Added logging to DRS actions

Asura5:
## Link to JIRA ticket

DRS possible timeouts cannot be diagnosed due to poor logging on the DRS Service.

Added logging.

### DIFF
--- a/RepairsApi/V2/UseCase/CreateWorkOrderUseCase.cs
+++ b/RepairsApi/V2/UseCase/CreateWorkOrderUseCase.cs
@@ -52,35 +52,38 @@ namespace RepairsApi.V2.UseCase
 
         public async Task<CreateOrderResult> Execute(WorkOrder workOrder)
         {
-            ValidateRequest(workOrder);
-            AttachUserInformation(workOrder);
-            workOrder.DateRaised = DateTime.UtcNow;
-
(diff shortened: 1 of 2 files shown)


RepairsApi/V2/UseCase/CreateWorkOrderUseCase.cs
```diff
@@ -65,15 +65,21 @@ public async Task<CreateOrderResult> Execute(WorkOrder workOrder)
             var notification = await NotifyHandlers(workOrder);
             var result = new CreateOrderResult(id, workOrder.StatusCode, workOrder.GetStatus());
```
kylealexanderhackney, line 66: WE can put using var scope = _logger.BeginScope(Guid.NewGuid()); at the top of thi method and it would mean any logs made as part of this use case and be extension any notifications fired will contain a guid that can be used to link the execution in the log

kylealexanderhackney (approved):
One suggestion but not necessary

RepairsApi/V2/UseCase/CreateWorkOrderUseCase.cs
Asura5: Ah nice - IDisposable so reference implementation is via a `using` - have updated.

PR #276, opened 2021-06-07
Assign operatives

kylealexanderhackney:
Adds the schema and functionality for assigning operatives to work orders in a many to many fashion.
The pr has gotten larger as the addition of many to many causes circular reference so the object generators require changing

- [ ] - Migration
- [x] - Check Coverage
- [x] - Integration tests


RepairsApi/V2/Infrastructure/Operative.cs
```diff
@@ -5,13 +5,16 @@
 
 namespace RepairsApi.V2.Infrastructure
 {
+    [Index(nameof(PayrollNumber), IsUnique = true)]
```
josh-dadswell-hackney, line 8: remove index as payroll is not unique

josh-dadswell-hackney (approved):

PR #288, opened 2021-06-11
Assign drs operative

josh-dadswell-hackney:
Removed code I suspect was accidentally committed which added cautionary alerts to the order comments field, meaning that iof the user maxed out the comments field it would fail to create the order in DRS


RepairsApi/V2/Services/DRS/DrsMapping.cs
```diff
@@ -56,9 +56,7 @@ private async Task<order> CreateOrder(WorkOrder workOrder)
             var locationAlerts = property != null ? await _alertsGateway.GetLocationAlertsAsync(property.PropertyReference) : null;
             var tenureInfo = property != null ? await _tenancyGateway.GetTenancyInformationAsync(property.PropertyReference) : null;
             var personAlerts = tenureInfo != null ? await _alertsGateway.GetPersonAlertsAsync(tenureInfo.TenancyAgreementReference) : null;
-            var orderComments = $"Property Alerts {locationAlerts?.Alerts.ToDescriptionString()} " +
```
josh-dadswell-hackney, line 59: Believe this was committed by accident

kylealexanderhackney (approved):

PR #291, opened 2021-06-15
Restrict returned work orders for DLO operatives

pixeltrix:
DLO operatives can raise repairs so rather than create a new role we give them both agent and contractor permissions.

### DIFF
--- a/RepairsApi/V2/Filtering/WorkOrderFilterProvider.cs
+++ b/RepairsApi/V2/Filtering/WorkOrderFilterProvider.cs
@@ -43,14 +43,22 @@ namespace RepairsApi.V2.Filtering
             var liveContractors = await _scheduleOfRatesGateway.GetLiveContractors();
             var filterOptions = new List<FilterOption>(liveContractors.Select(c => new FilterOption { Key = c.ContractorReference, Description = c.ContractorName }));
 
-            if (!_currentUserService.HasAnyGroup(UserGroups.Agent, UserGroups.AuthorisationManager, UserGroups.ContractManager))
+            if (_currentUserService.HasAnyGroup(UserGroups.AuthorisationManager, UserGroups.ContractManager))
             {
+                // Authorisation and contract managers can see all work orders
+                return filterOptions;
+            }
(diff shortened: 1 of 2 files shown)


pixeltrix:
Required for https://github.com/LBHackney-IT/repairs-hub-frontend/pull/228

RepairsApi/V2/Filtering/WorkOrderFilterProvider.cs
```diff
@@ -43,14 +43,22 @@ private async Task<List<FilterOption>> GetContractors()
             var liveContractors = await _scheduleOfRatesGateway.GetLiveContractors();
             var filterOptions = new List<FilterOption>(liveContractors.Select(c => new FilterOption { Key = c.ContractorReference, Description = c.ContractorName }));
 
-            if (!_currentUserService.HasAnyGroup(UserGroups.Agent, UserGroups.AuthorisationManager, UserGroups.ContractManager))
+            if (_currentUserService.HasAnyGroup(UserGroups.AuthorisationManager, UserGroups.ContractManager))
             {
+                // Authorisation and contract managers can see all work orders
```
kylealexanderhackney, line 48: This restriction will break the flow of the system as users will be able to raise orders for contractors and not have permission to see them

RepairsApi/V2/Filtering/WorkOrderFilterProvider.cs
```diff
@@ -43,14 +43,22 @@ private async Task<List<FilterOption>> GetContractors()
...
+                return filterOptions;
+            }
+            else if (_currentUserService.HasGroup(UserGroups.Agent) && !_currentUserService.HasGroup(UserGroups.Contractor))
```
kylealexanderhackney, line 51: group membership should always add new permissions never subtract existing one or we open ourselves up to a lot of security edge cases and complexity. 
this seems more like a new user type needs to be added

kylealexanderhackney (changes requested):
I maybe missing something but what do these changes have to do with operatives?

pixeltrix:
Doing this in the front end via filters

PR #296, opened 2021-06-16
only send url if token is present

josh-dadswell-hackney:


RepairsApi/V2/Factories/ResponseFactory.cs
```diff
@@ -127,11 +127,11 @@ public static WorkOrderResponse ToResponse(this Infrastructure.WorkOrder workOrd
         {
             Infrastructure.PropertyClass propertyClass = workOrder.Site?.PropertyClass?.FirstOrDefault();
             string addressLine = propertyClass?.Address?.AddressLine;
-            var managementUri = new UriBuilder(drsManagementAddress)
+            var managementUri = workOrder.ExternalSchedulerReference is null ? null : new UriBuilder(drsManagementAddress)
             {
                 Port = -1,
                 Query = $"tokenId={workOrder.ExternalSchedulerReference}"
-            };
+            }.Uri;
```
kylealexanderhackney, line 134: can we put the .Uri on line 165 IMO that reads a bit cleaner

RepairsApi/V2/Factories/ResponseFactory.cs
```diff
@@ -127,11 +127,11 @@ public static WorkOrderResponse ToResponse(this Infrastructure.WorkOrder workOrd
         {
             Infrastructure.PropertyClass propertyClass = workOrder.Site?.PropertyClass?.FirstOrDefault();
             string addressLine = propertyClass?.Address?.AddressLine;
-            var managementUri = new UriBuilder(drsManagementAddress)
+            var managementUri = workOrder.ExternalSchedulerReference is null ? null : new UriBuilder(drsManagementAddress)
```
kylealexanderhackney, line 130: could this go into a method or property/extension on workorder. It feels like it would be cleaner

kylealexanderhackney (approved):

RepairsApi/V2/Factories/ResponseFactory.cs
josh-dadswell-hackney: No because it could be null by that point

PR #308, opened 2021-06-23
Add note to WO when drs update received

josh-dadswell-hackney:
https://www.pivotaltracker.com/story/show/178480897


RepairsApi.Tests/V2/Services/DRS/BackgroundService/DrsBackgroundServiceTests.cs
```diff
@@ -126,6 +132,24 @@ await testFunc.Should().ThrowAsync<ResourceNotFoundException>()
             _loggerMock.VerifyLog(LogLevel.Error);
         }
 
+        [Test]
+        public async Task AddsNote()
```
Asura5, line 136: Any paths that don't create a note? Would it worth adding a DoesNotAddNote test?

Asura5 (approved):

PR #310, opened 2021-06-23
allow multiple sor codes to be sent to drs

josh-dadswell-hackney:
https://www.pivotaltracker.com/story/show/178640055

### DIFF
--- a/RepairsApi.Tests/V2/Services/DRS/DrsMappingTests.cs
+++ b/RepairsApi.Tests/V2/Services/DRS/DrsMappingTests.cs
@@ -262,12 +262,13 @@ namespace RepairsApi.Tests.V2.Services
 
         private IList<ScheduleOfRatesModel> SetupSorCodes(WorkOrder workOrder)
         {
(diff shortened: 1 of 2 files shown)


kylealexanderhackney (approved):

RepairsApi.Tests/V2/Services/DRS/DrsMappingTests.cs
```diff
@@ -262,12 +262,13 @@ public async Task MapsAlerts()
 
         private IList<ScheduleOfRatesModel> SetupSorCodes(WorkOrder workOrder)
         {
-            var sorCodes = workOrder.WorkElements.FirstOrDefault()?.RateScheduleItem
-                .Select(rsi => new ScheduleOfRatesModel
+            var sorCodes = workOrder.WorkElements.SelectMany(we => we.RateScheduleItem.Select(rsi => new { rsi, we }))
```
Asura5, line 265: Just check that this change isn't going to make the null check on line 274 redundant. AFAIK the SelectMany will always return an IEnumerable<TSource>

PR #341, opened 2021-07-07
Added feature filter based on user google groups

kylealexanderhackney:
This works adds a custom feature filter that can be configured with a list of google groups. features set to use the filter will be available if the user passed is in one of the groups.

It includes a modification to the integration tests that enforces operative assignment for a specific group as a demo


RepairsApi/GroupFeatureFilter.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class GroupFeatureFilter : IFeatureFilter
+    {
+        private readonly ICurrentUserService _currentUserService;
+
+        public GroupFeatureFilter(ICurrentUserService currentUserService)
+        {
+            _currentUserService = currentUserService;
+        }
+
+        public Task<bool> EvaluateAsync(FeatureFilterEvaluationContext context)
```
Asura5, line 27: Unclear method naming  - could make it more explicit - DoesUserHaveFeatures etc

Asura5 (approved):

RepairsApi/GroupFeatureFilter.cs
kylealexanderhackney: Its the interface method so we dont control the name

PR #362, opened 2021-08-06
Introduce work order statuses to DB

Asura5:
## Link to JIRA ticket

First part of https://www.pivotaltracker.com/n/projects/2500129/stories/179054706

## Describe this PR

The problem is 502/504 errors being generated by the API gateway. This is due to the lambda timout which is in turn due to the generated SQL query taking longer than the 15 second limit. Looking at the raw SQL produced by the LINQ, the query would currently take approx 4 mins. This is because the entire workorder table is being retrieved from the DB and sorted in memory.

An alternative approach is to sort, and skip/take at the database layer. One impediment to this (and the reason that I believe it has not been done this way so far) is because the sorting is based on a non-linear key (i.e. it cannot be translated via LINQ-to-SQL). Therefore, we introduce work_order_status with a sortable int display priority.


LBHBomotoso (approved):

RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
```diff
@@ -1598,6 +1598,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                    .HasColumnType("integer")
+                    .HasColumnName("id")
+                    .UseIdentityByDefaultColumn();
+
+                wos.Property<string>("Description")
+                    .HasColumnType("string")
+                    .HasColumnName("status_description");
+
+                wos.Property<string>("DisplayPriority")
+                    .HasColumnType("string")
```
NeilvB, line 1614: How come this has generated a string column on the model when the migration specified an `int`?

Asura5: Good question - updated to integer

PR #364, opened 2021-08-09
Exclude migrated wo from search

Asura5:
Exclude the historical migration via a parameter in to the work order search. Historical data is not needed in many contexts, and excluding it speeds up query execution substantially


LBHBomotoso (approved):

RepairsApi/V2/Gateways/RepairsGateway.cs
```diff
@@ -37,12 +37,19 @@ public Task<IEnumerable<WorkOrder>> GetWorkOrders(params Expression<Func<WorkOrd
         }
 
         public async Task<IEnumerable<WorkOrder>> GetWorkOrders(IFilter<WorkOrder> filter)
+        {
+            return await GetWorkOrders(filter, true);
```
josh-dadswell-hackney, line 41: Could the same thing not be done using a default param reduces code

josh-dadswell-hackney (approved):

PR #383, opened 2021-09-01
Add updated time to tasks

josh-dadswell-hackney:
https://www.pivotaltracker.com/story/show/178775488


RepairsApi/V2/Infrastructure/RateScheduleItem.cs
```diff
@@ -12,6 +12,7 @@ public class RateScheduleItem
         public string CustomName { get; set; }
         public virtual Quantity Quantity { get; set; }
         public DateTime? DateCreated { get; set; }
+        public DateTime? DateUpdated { get; set; }
```
Asura5, line 15: Is there a DB migration for this field?

josh-dadswell-hackney: good catch, on this note I have been wondering if we should implement a check in the CI for missed migrations, should be simple - run migrations add check its an empty migration. Just no idea how to do iit in circle

josh-dadswell-hackney: Added migration

Asura5 (approved):

PR #392, opened 2021-09-06
Check migrations

josh-dadswell-hackney:
Added into Build_Test step a checkl for blank migration


RepairsApi.Tests/check-migrations.sh
```diff
@@ -0,0 +1,28 @@
...
+
+printf 'Creating test migration\n'
+dotnet ef migrations add blank --project RepairsApi > /dev/null
+
+migrationFile=$(find ./RepairsApi/V2/Infrastructure/Migrations/*blank.cs)
+
+filesize=$(stat -c%s "$migrationFile")
+maxsize=346
+if (( filesize > maxsize )); then
+    printf 'Missing migration:\n'
```
Asura5, line 16: Make it more obvious that this is a failure state

RepairsApi.Tests/check-migrations.sh
```diff
@@ -0,0 +1,28 @@
...
+    cat $migrationFile
+    printf 'Migration is not blank - Missing migration detected!!!!\nPlease check above to see the generated migration\n'
+    failed=1
+else
+    printf 'Migration is blank - No misssing migrations found :)\n'
```
Asura5, line 21: Typo "misssing"

RepairsApi.Tests/check-migrations.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+printf 'Removing generated migration\n'
+dotnet ef migrations remove --project RepairsApi -f > /dev/null
```
Asura5, line 25: The only worry I have about this is if for some reason the migration add fails, when this runs it could remove something from a previous PR

RepairsApi.Tests/check-migrations.sh
josh-dadswell-hackney: What would you recommend?

RepairsApi.Tests/check-migrations.sh
josh-dadswell-hackney: Not a real issue as this is runing on a cloned copy of the code anyway, removing the migration isnt 100% necessary as the whole container is trashed after it runs.

RepairsApi.Tests/check-migrations.sh
Asura5: FAILED: Missing migration:

Asura5: It's just more helpful for people who are searching the log for failures

josh-dadswell-hackney: Sure makes sense

Asura5 (approved):

PR #403, opened 2021-09-22
Set booking start end time

josh-dadswell-hackney:


RepairsApi.Tests/V2/Services/DRS/DrsMappingTests.cs
```diff
@@ -422,6 +421,11 @@ private static void VerifyCompleteOrder(updateBooking updateBooking, WorkOrder w
                 .Excluding(o => o.theBusinessData)
             );
             updateBooking.updateBooking1.theBooking.theOrder.status.Should().Be(orderStatus.COMPLETED);
+
+            var expectedStart = workOrder.ClosedDate - (booking.assignedEnd - booking.assignedStart) ?? booking.assignedStart;
+            var expectedEnd = workOrder.ClosedDate ?? booking.assignedEnd;
```
LBHBomotoso, line 426: Isn't this in the usecase?

josh-dadswell-hackney: No the use case fires a completion event which is picked up by the DRSNotificationHandler, this change is making sure the end time of the booking is whatever was selected by the user when they closed the work order. The start time is then shifted by the duration.

josh-dadswell-hackney: This is specifically about building the SOAP request for sending to DRS

LBHBomotoso: cool, understood

LBHBomotoso (approved):

PR #409, opened 2021-09-29
Working version with tests

josh-dadswell-hackney:


RepairsApi/Handler.cs
```diff
@@ -1,58 +1,83 @@
...
+    }
 
-        public static async Task SyncDRSWorkOrders()
+
+    public static class ConfigureServices
+    {
+        public static void Configure(this IServiceCollection services)
         {
-            var webHost = BuildWebHost(null);
+            var connectionString = Environment.GetEnvironmentVariable("CONNECTION_STRING");
```
Asura5, line 64: Could include a null check here to handle if the variable cannot be found

RepairsApi/aws-lambda-tools-defaults.json
```diff
@@ -0,0 +1,19 @@
+{
```
Asura5, line 1: Is this file included to help others run the lambda locally?

josh-dadswell-hackney: yes

RepairsApi/Handler.cs
josh-dadswell-hackney: This is unused I just forgot to remove it.

Asura5 (approved):

PR #415, opened 2021-10-05
disable pagination of orders when syncing

josh-dadswell-hackney:


RepairsApi/V2/UseCase/ListWorkOrdersUseCase.cs
```diff
@@ -33,18 +33,21 @@ public async Task<IEnumerable<WorkOrder>> Execute(WorkOrderSearchParameters sear
 
             workOrders = EnsureSorting(searchParameters, workOrders);
 
+            if (searchParameters.PageSize <= 0)
+            {
+                return workOrders;
```
Asura5, line 38: Are we not going to hit the lambda execution time limit with this eventually?

Asura5: Also, I think anything <0 should be an error state, 0 exactly being this code path

josh-dadswell-hackney: Hopefully not, but potentially, depends on how many work orders we end up raising in one day. How else would you recommend we handle it?

Asura5: Let's cross that bridge when we come to it

Asura5 (approved):

PR #434, opened 2021-11-02
Create separate operative response model for work orders

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180152765

## Describe this PR

Created a WorkOrderOperativeResponse model which inherits fields from OperativeResponse and introduces a job percentage property.


RepairsApi/V2/Factories/ResponseFactory.cs
```diff
@@ -322,9 +322,9 @@ public static OperativeResponse ToResponse(this Infrastructure.Operative operati
             };
         }
 
-        public static OperativeResponse ToResponse(this Infrastructure.WorkOrderOperative woo)
+        public static WorkOrderOperativeResponse ToResponse(this Infrastructure.WorkOrderOperative woo)
```
Asura5, line 325: Is this going to return a `WorkOrderOperativeResponse` into places where the variables are declared as `OperativeResponse`? There will not be any compiler error thrown because a `WorkOrderOperativeResponse` is a valid `OperativeResponse` but you may still have the field

teodor-hristoforov-hackney: The overload should return a WorkOrderOperativeResponse when it's called with WorkOrderOperative as an argument and we expect it to return an OperativeResponse when the argument is of type Operative. I haven't implemented any other contingencies at the moment.

Asura5 (approved):

PR #437, opened 2021-11-03
Sor codes raisable flag

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180164676

## Describe this PR

Create new migration to add a "isRaisable" field to the sor codes table and use the field to filter when getting sor codes. The "Enabled" field will be removed in a subsequent PR as it is now redundant.


RepairsApi/V2/Controllers/ScheduleOfRatesController.cs
```diff
@@ -49,9 +49,9 @@ IScheduleOfRatesGateway scheduleOfRatesGateway
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
         [ProducesDefaultResponseType]
         [Authorize(Roles = UserGroups.Agent + "," + UserGroups.ContractManager + "," + UserGroups.AuthorisationManager + "," + UserGroups.Operative)]
-        public async Task<IActionResult> ListSorCodes([FromQuery][Required] string tradeCode, [FromQuery][Required] string propertyReference, [FromQuery][Required] string contractorReference)
+        public async Task<IActionResult> ListSorCodes([FromQuery][Required] string tradeCode, [FromQuery][Required] string propertyReference, [FromQuery][Required] string contractorReference, [FromQuery][Required] bool isRaisable)
```
teodor-hristoforov-hackney, line 52: I can set this parameter as optional and default it to false if necessary

Asura5: Can you please clarify what isRaisable means? if it's false, true or null?

teodor-hristoforov-hackney: At the moment when when using this api it returns sor codes for jobs that are already raised (enabled flag). What this change is meant to do is pass another parameter into the api call to specify if you only want all the sor codes or only the ones that have a raised job.

teodor-hristoforov-hackney: This first PR introduces isRaisable which is the replacement of the Enabled flag. Once this PR is merged and tested we can create a subsequent PR to remove Enabled from the database and codebase.

Asura5: so is isRaisable = false going to return all?  isRaisable = true going to return all? What will the null behaviour be?
IMHO this has make it _less_ descriptive, not more.
You need to have this clear from both backend and frontend before moving forwards

Asura5 (approved):

RepairsApi/V2/Controllers/ScheduleOfRatesController.cs
siapankina: Sorry to intervene :) @Asura5  The scenario is:
-when user raises a repair then with GET request will send params: 
 `isRaisable=true` (we need to get only SORs that has flag `isRaisable:true`)
 - when user does variation f/e sends GET request without `isRaisable` (need to to get all SORs)

PR #438, opened 2021-11-04
Minor change to the way isRaisable flag is being used when getting sor codes

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180164676

## Describe this PR
Change to the isRaisable flag so when it is passed as null it returns all sor codes, else it returns the codes based on the isRaisable value


RepairsApi.Tests/V2/Gateways/ScheduleOfRatesGatewayTests.cs
```diff
@@ -206,12 +206,51 @@ public async Task ReturnsAllIfIsRaisableIsFalse()
             var expectedTrade = await SeedTrade(Guid.NewGuid().ToString());
             var validContracts = await SeedContracts(expectedProperty, DateTime.UtcNow.AddDays(-7), DateTime.UtcNow.AddDays(7), contractorReference);
 
+            await SeedSorCodes(expectedPriority, expectedProperty, expectedTrade, validContracts.First());
+            var expectedCodes = await SeedSorCodes(expectedPriority, expectedProperty, expectedTrade, validContracts.First(), false, false);
+
+            await InMemoryDb.Instance.SaveChangesAsync();
+
+            await GetAndValidateCodes(expectedProperty, expectedTrade, contractorReference, expectedCodes, "false");
```
Asura5, line 214: Is there any way we can move the boolean up the call stack at all? This string "false" business is really hacky....

teodor-hristoforov-hackney: Is this what you had in mind?

PR #450, opened 2021-11-10
RH - post message on work order completion

Asura5:
Please see the listener PR for dev notes & caveats


RepairsApi/V2/Boundary/OutboundQueue/WorkOrderOperativeSmvData.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+
+namespace RepairsApi.V2.Boundary.OutboundQueue
+{
+    public class WorkOrderOperativeSmvData
+    {
+        public string WorkOrderId { get; set; }
+        public string Address { get; set; }
```
pixeltrix, line 8: The designs for the bonus calculation app call for a description of the job as well - shouldn't it  be included here?

RepairsApi/V2/Services/RepairsSnsService.cs
```diff
@@ -0,0 +1,92 @@
...
+                    },
+                    EventData = new WorkOrderOperativeSmvData
+                    {
+                        WorkOrderId = workOrder.Id.ToString(),
+                        Address = workOrder.LocationOfRepair,
+                        StandardMinuteValue = totalSMV,
+                        OperativeId = operative.OperativeId.ToString(),
+                        JobPercentage = operative.JobPercentage,
+                        ClosedTime = workOrder.WorkOrderComplete.ClosedTime
+                    }
```
pixeltrix, line 85: By publishing the percentage / total SMV as individual messages we're going to lose some SMV in rounding, e.g. if a job has 1000 SMV and each operative gets a third then we'll lose 0.0001 SMV since it's stored at 4 decimal places in the bonus calculation app. This shouldn't be a problem since there's no reconciliation between the two systems but we should be aware of it.

RepairsApi/V2/Boundary/OutboundQueue/WorkOrderOperativeSmvData.cs
Asura5: Added. In the listener, have mapped to the Comment field on PayElement as that seemed the most appropriate

RepairsApi/V2/Services/RepairsSnsService.cs
Asura5: Granted - but wouldn't we lose that accuracy anyway when we split it in the BC listener? Perhaps I am misunderstanding, but if a single row in the pay_elements table maps to a single operative (which it is, through timesheets to operative_id) then we either lose the accuracy in message creation or when it's stored, but we lose it either way...

Asura5: Or are you saying we could apply rounding to one of the operatives etc

pixeltrix: You'd do operative 1 as 33.3333%, operative 2 as 33.3333% and operative 3 as 33.3334% so that it always adds up to 100% but I don't think it's worth worrying about as it's 1/10000 of a minute or 0.006 seconds.

Asura5: Especially as somebody would have to tell the operatives that one of them is getting an extra 0.006 seconds - I can only imagine what a debate that would be

pixeltrix (approved):
Not necessary now, but to stop the 'OMG, you've leaked a secret' messages on other projects I've generated them either as part of CI configuration or in the one-time setup phase of whatever language/testing framework I'm using. This is what I did for the Petitions website: https://github.com/alphagov/e-petitions/commit/72006c11b4e4ff2e1a317b2c949d11956035824c

RepairsApi/V2/Services/RepairsSnsService.cs
```diff
@@ -0,0 +1,91 @@
...
+                    DateTime = DateTime.UtcNow,
+                    User = new User
+                    {
+                        Name = user.Claims.First(c => c.Type == ClaimTypes.Name).Value,
+                        Email = user.Claims.First(c => c.Type == ClaimTypes.Email).Value
+                    },
+                    EventData = new WorkOrderOperativeSmvData
+                    {
+                        WorkOrderId = workOrder.Id.ToString(),
+                        Address = workOrder.LocationOfRepair,
```
pixeltrix, line 78: Is there anything in `LocationOfRepair`? I can't see anything in the database.

Asura5: Correct - changed. It's actually in workOrder.Site.PropertyClass[].Address.AddressLine. We should probably delete that column as people may trip up in future.

RepairsApi/V2/Services/RepairsSnsService.cs
```diff
@@ -0,0 +1,91 @@
...
+                        Name = user.Claims.First(c => c.Type == ClaimTypes.Name).Value,
+                        Email = user.Claims.First(c => c.Type == ClaimTypes.Email).Value
+                    },
+                    EventData = new WorkOrderOperativeSmvData
+                    {
+                        WorkOrderId = workOrder.Id.ToString(),
+                        Address = workOrder.Site.PropertyClass.FirstOrDefault().Address.AddressLine,
+                        Description = workOrder.DescriptionOfWork,
+                        StandardMinuteValue = totalSMV,
+                        OperativeId = operative.OperativeId.ToString(),
```
pixeltrix, line 81: @Asura5 something I only realised today when importing the data - `OperativeId` here is a generated id in the database and not the payroll number 😬

Asura5: Added to new PR

PR #454, opened 2021-11-17
Add OOH flag, op cost and income columns to sor_codes

Asura5:
- Add out of hours to sor_codes table
- Add operative cost and income to sor_codes table


LBHCallumM (approved):
Looks good to me.

PR #467, opened 2021-11-25
Stop adding prefixes to empty comments

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180264573

## Describe this PR

When updating job status we will no longer add prefixes if the comment is empty which would prevent us from creating comments such as "Variation Reason: "


LBHCallumM (commented):
Is it worth adding tests for this?

teodor-hristoforov-hackney:
> Is it worth adding tests for this?

You're right, it's best practice to add testing. I'll add some.

RepairsApi/V2/Helpers/JobStatusUpdateExtensions.cs
```diff
@@ -6,7 +6,7 @@ public static class JobStatusUpdateExtensions
     {
         public static void PrefixComments(this JobStatusUpdate jobStatusUpdate, string prefix)
         {
-            if (jobStatusUpdate.Comments is null || !jobStatusUpdate.Comments.StartsWith(prefix))
+            if (jobStatusUpdate.Comments != null && !jobStatusUpdate.Comments.StartsWith(prefix))
```
Asura5, line 9: But this means we will have empty job status updates in the Notes tab. I think @NeilvB and @siapankina were looking at this. See this thread: [https://hackit-lbh.slack.com/archives/C01EG9951L1/p1637594810081600](https://hackit-lbh.slack.com/archives/C01EG9951L1/p1637594810081600)

teodor-hristoforov-hackney: @Asura5 It's following your second suggestion in the thread "Remove the prefix for blank comments and change the frontend to not display those items". It will need some front end work as well, but so would the other options.

NeilvB: I am confident that the backend already filters out empty JobStatusUpdate notes from the notes endpoint

https://github.com/LBHackney-IT/repairs-api-dotnet/blob/1d546c89c0c7e2da7d157674aee5758394150aea/RepairsApi/V2/UseCase/ListWorkOrderNotesUseCase.cs#L27

Asura5 (approved):

PR #482, opened 2021-12-01
Fix issue where work order closed date does not match when work order was set to close

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/178794452

## Describe this PR

When completing work order the date should be the same as when the operative closed the work order.


RepairsApi/V2/UseCase/CompleteWorkOrderUseCase.cs
```diff
@@ -107,8 +107,6 @@ private async Task UpdateWorkOrderStatus(WorkOrder workOrder, WorkOrderComplete
                     break;
                 default: throw new NotSupportedException(Resources.UnsupportedWorkOrderUpdate);
             }
-
-            workOrder.ClosedDate = workOrderComplete.ClosedTime ?? update.EventTime;
```
Asura5, line 111: I think this is going to cause a lot of issues with the ClosedDate not being set - can we talk about this please?

PR #487, opened 2021-12-03
Adds option to split work orders using JobStatusUpdate endpoint

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180262836

## Describe this PR

Adds an optional "isSplit" parameter to JobStatusUpdate api that updates the split job status of work orders.
If the field is not provided the work order keeps the previous value it was assigned.


RepairsApi/V2/Infrastructure/JobStatusUpdate.cs
```diff
@@ -20,6 +20,7 @@ public class JobStatusUpdate
         public virtual AdditionalWork AdditionalWork { get; set; }
         public virtual WorkOrder RelatedWorkOrder { get; set; }
         public int? RelatedWorkOrderId { get; set; }
+        public bool? IsSplit { get; set; }
```
LBHCallumM, line 23: Surely this can only be true or false. Does it need to be nullable?

teodor-hristoforov-hackney: It's an optional parameter. This field is only used as a trail of what the user has updated on each job.

The place where we actually check the value is within the Work Order which has a field with the same name.

The 3 use cases here are:
* The user sets it to true -> work order is_split is set to true.
* The user sets it to false -> work order is_split is set to false.
* The user does not provide the parameter -> work order is_split stays the same value as it was before

RepairsApi/V2/Generated/Extensions/JobStatusUpdateExtensions.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace RepairsApi.V2.Generated
+{
+    public partial class JobStatusUpdate
+    {
+        [Newtonsoft.Json.JsonProperty("IsSplit", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
```
Asura5, line 10: DisallowNull but type is nullable?

pixeltrix: > DisallowNull but type is nullable?

The discussion we had was that `is_split` on `job_status_updates` acts as a record of what was in the request. The request should either be able to change work order from non-split to split or vice-versa. That's why `DisallowNull` is set here so that you can either omit it or change the state of the work order.

Asura5 (approved):

RepairsApi/V2/Generated/Extensions/JobStatusUpdateExtensions.cs
Asura5: Ah - seems logical based on that context

PR #513, opened 2021-12-10
Integrate GetProperty endpoint with new Tenure Information API

LBHCallumM:
## Describe this PR

Integrate Tenure Information API with GetProperty endpoint.

I have included a Feature Toggle to prevent other deployments from being blocked.

The existing TenancyGateway has been renamed to UHTenancyGateway to make it more obvious what is does. The new gateway is called TenureInformationGateway.

The following methods also use the UHTenancyGateway . I will create separate PR's to update these:
- ListAlertsUseCase
- DrsMapping

#### _Checklist_


Asura5:
When you do ListAlertsUseCase, this whole path can be stripped out of the codebase as we are no longer going to get alerts from UH, we are going to reply exclusively on the sheet

Asura5 (approved):
I'm just wondering if there's going to be any issues with using the new tenure API for some use cases but not everything - a moot point perhaps as we're already live with it but something to look out for

LBHCallumM:
> I'm just wondering if there's going to be any issues with using the new tenure API for some use cases but not everything - a moot point perhaps as we're already live with it but something to look out for

I will update the other usecases before this goes live. I just didnt want the PR to be too big.

PR #530, opened 2021-12-20
Tiny - add BackupPolicy tag to RDS (Hackney devops centralized backup trigger)

Asura5:
The "BackupPolicy" is an agreed flag for the automatic AWS Backup pickup - just adding to dev first because I'm fairly sure it will nuke the other flags on this resource, in which case I'll add them all - OR there's some shared config/tags in the common module that I'm not aware of


Asura5:
For my ref - tags before:

  | Environment | development
  | Name | repairs_db-db-development
  | project_name | repairs hub
  | terraform-managed | true

PR #533, opened 2021-12-23
Add LastDrsSync column to workOrder

LBHCallumM:
#### _Checklist_


RepairsApi/V2/Gateways/RepairsGateway.cs
```diff
@@ -151,5 +151,15 @@ public Task SaveChangesAsync()
         {
             return _repairsContext.SaveChangesAsync();
         }
+
+        public async Task UpdateLastDrsSyncTime(int workOrderId)
```
Asura5, line 155: The issue with doing this way is that it will impact performance in a significant way - an extra get and store operation for every update. Instead of int workOrderID, take a WorkOrder object and set the column, without saving changes. Refactor  this along with UpdateWorkOrderPlannerComments so that there's only one call to Get the order, and one SaveChanges, to minimise load of the DB

LBHCallumM: Makes sense. I have combined both of these methods.

RepairsApi/V2/Gateways/RepairsGateway.cs
```diff
@@ -140,10 +140,14 @@ public async Task UpdateWorkOrderStatus(int workOrderId, WorkStatusCode newCode)
             await _repairsContext.SaveChangesAsync();
         }
 
-        public async Task UpdateWorkOrderPlannerComments(int workOrderId, string plannerComments)
+        public async Task UpdateWorkOrderFromDrsSync(int workOrderId, string plannerComments)
```
Asura5, line 143: UpdateWorkOrderAttributes

Asura5 (approved):

PR #560, opened 2022-01-18
Fix sync failure for DRS "general" resource groups

Asura5:
https://www.pivotaltracker.com/n/projects/2500129/stories/180913479

The sync is ending early after encountering an error - e.g. [this trace](https://eu-west-2.console.aws.amazon.com/cloudwatch/home?region=eu-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Frepairs-api-production-sync-drs/log-events/2022$252F01$252F17$252F$255B$2524LATEST$255D917ef9931fae48c78e0b33262f4f9df4$3Fstart$3D1642436143560$26refEventId$3D36627549944618897772986262959767467684349309218349187076)

It's due to a NRE from the operatives LINQ statement - if a particular operative cannot be found, the operatives gateway will return null and the select will try to operate on this. This can happen with "operatives" who are actually groups (e..g for this workorder, it's HH Gas Breakdown (0202271) - but we don't have an operative called HH Gas Breakdown - as far as mobile working is concerned, all operatives are individuals


RepairsApi/V2/Services/DRS/DrsService.cs
```diff
@@ -206,11 +207,21 @@ public async Task UpdateWorkOrderDetails(int workOrderId)
...
+                var jobPercentage = 100.0 / operatives.Length;
+
+                _logger.LogInformation($"Updating operatives [{string.Join(", ", operativePayrollIds)}] to jobPercentage {jobPercentage} for workOrderId: {workOrderId}");
+
+                await _operativesGateway.AssignOperatives(workOrderId, OperativeAssignmentType.Automatic, operatives.Select(o => (o.Id, jobPercentage)).ToArray());
+            }
+            else
+            {
+                _logger.LogInformation($"Gateway did not return any operatives for workorderId: {workOrderId}, though {string.Join(", ", operativePayrollIds)} were searched");
+            }
```
LBHCallumM, line 223: Could this be extracted into its own method? UpdateWorkOrderDetails is already quite large.

RepairsApi/V2/Services/DRS/DrsService.cs
```diff
@@ -206,11 +207,21 @@ public async Task UpdateWorkOrderDetails(int workOrderId)
             var operativePayrollIds = theResources.Select(r => r.externalResourceCode);
             var operatives = await Task.WhenAll(operativePayrollIds.Select(i => _operativesGateway.GetAsync(i)));
 
-            var jobPercentage = 100.0 / operatives.Length;
+            operatives = operatives.Where(o => o != null).ToArray();
```
LBHCallumM, line 210: Could getting the operatives list be a single method?

LBHCallumM (approved):

teodor-hristoforov-hackney (approved):

RepairsApi/V2/Services/DRS/DrsService.cs
Asura5: Good suggestion - changed

RepairsApi/V2/Services/DRS/DrsService.cs
Asura5: Good suggestion - changed

PR #563, opened 2022-01-19
Don't allow assignment of more/less than ~100% job coverage

Asura5:
This is intended as a safety net where methods are calling the Operatives Gateway to assign bizarre job update and completion amounts. These will therefore be pushed upstream.

There is no valid reason why a work order can have != 100% assigned to it (with the obvious precision caveat of 3 operatives dividing a job etc...), so this will stop that from happening.

The tests have been refactored as they were clearly wrong, and the logging has been improved


RepairsApi/V2/Gateways/OperativesGateway.cs
```diff
@@ -63,17 +66,30 @@ public async Task ArchiveAsync(string operativePrn)
 
         public async Task AssignOperatives(int workOrderId, OperativeAssignmentType type, params (int id, double jobPercentage)[] operatives)
         {
-            var workOrderOperatives = _context.WorkOrderOperatives.Where(woo => woo.WorkOrderId == workOrderId).AsEnumerable();
-            _context.WorkOrderOperatives.RemoveRange(workOrderOperatives);
-            var assignments = operatives.Select(o => new WorkOrderOperative
+            var total = operatives.Sum(op => op.jobPercentage);
+
+            if (total < 99 || total > 100) // Allow for 3-way split (will be 99.9999%)
```
pixeltrix, line 71: This could be written as this: 

``` c#
 if (Math.Abs(100 - total) > 0.0001)
```

pixeltrix (approved):
Math for suggestion only

LBHCallumM (approved):

teodor-hristoforov-hackney (approved):

RepairsApi/V2/Gateways/OperativesGateway.cs
Asura5: Good idea - changed

PR #565, opened 2022-01-20
Cloudwatch Alerts for DRS Sync - automatic notification for systematic failures

Asura5:
Theme: "Knowing our systems are going wrong before our customers know they are going wrong"

- Currently we do not have any automated alerts for our critical processes in place - especially our "hidden" (background) processes
- This change implements a Cloudwatch alarm to automatically notify users of a consistent failure due to NRE (currently the no1 transient failure mode for the DRS sync) - though this can be expanded easily by adjusting the filter pattern
- The accepted AWS pattern is that CloudWatch will push a notification to a queue which will than trigger (any number of things) but in this case an email alert to myself (though this is also configurable)


.circleci/config.yml
```diff
@@ -75,6 +75,9 @@ commands:
       - run:
           name: Install serverless CLI
           command: npm i -g serverless
+      - run:
+          name: Install AWs Alerts plugin
```
pixeltrix, line 79: Minor typo on `AWs`

RepairsApi/serverless.yml
```diff
@@ -179,3 +182,20 @@ custom:
...
         - subnet-0beb266003a56ca82
         - subnet-06a697d86a9b6ed01
+  alerts:
+    stages:
+      - production
+    topics:
+      alarm: ${ssm:/sns-topic/${self:provider.stage}/repairs/alarm/arn}
+      notifications:
+        - protocol: email
+          endpoint: david.dean@hackney.gov.uk
```
pixeltrix, line 192: Should this be a group? Not sure whether repairs-hub-support is the right one though.

terraform/production/main.tf
```diff
@@ -40,6 +41,58 @@ resource "aws_ssm_parameter" "repairs_sns_arn" {
...
+          "Action": "sqs:SendMessage",
+          "Resource": "${aws_sqs_queue.repairs_alarms_queue.arn}",
+          "Condition": {
+          "ArnEquals": {
+              "aws:SourceArn": "${data.aws_ssm_parameter.repairs_alarms_sns_arn.value}"
+          }
+          }
+      }
+      ]
+  }
```
pixeltrix, line 85: Indentation looks a bit off here - this is what I get after running it through a JSON formatter

``` JSON
{
    "Version": "2012-10-17",
    "Id": "sqspolicy",
    "Statement": [
        {
            "Sid": "First",
            "Effect": "Allow",
            "Principal": "*",
            "Action": "sqs:SendMessage",
            "Resource": "${aws_sqs_queue.repairs_alarms_queue.arn}",
            "Condition": {
                "ArnEquals": {
                    "aws:SourceArn": "${data.aws_ssm_parameter.repairs_alarms_sns_arn.value}"
                }
            }
        }
    ]
}
```

pixeltrix (approved):
Just a couple of formatting suggestions and a question about who should get the alert

teodor-hristoforov-hackney (approved):

PR #569, opened 2022-01-21
adding stored procedure to update is_raisable value for an SOR

elena-evans:
https://www.pivotaltracker.com/story/show/180994322

Adding a migration for a stored procedure to change the availability of an SOR code to be raised.
This replaces the need to use raw SQL in the database, and will be helpful for the back office application.


LBHCallumM (approved):
Is it possible to write tests for stored procedures? Might be worth considering.

teodor-hristoforov-hackney (approved):

teodor-hristoforov-hackney:
> Is it possible to write tests for stored procedures? Might be worth considering.

I agree that some testing should be put in place in the future. For now we have no examples for testing Entity Framework migrations and it would be difficult. One reason is that the in memory database we create for testing is fundamentally different than Postgres so we would need a new approach.

Asura5:
For future reference @elena-evans  @teodor-hristoforov-hackney  @LBHCallumM It's not really possible in raw SQL, but we could include it as part of the E2E tests, where we do use a Postgres adapater (MockWebApplicationFactory:80 - `UseNpgsql`). For example, in `ListOperativesE2ETests `we have `AddOperativesToDb`. This, at the moment, consists of manually filling the table. Howeer - it wouldn't be too difficult to replicate this test and change the implementation of `AddOperativesToDb `to run a stored proc. All that would be needed is to expose this procedure in the `RepairsContext`

PR #576, opened 2022-01-26
180890663/implement payment type property

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://www.pivotaltracker.com/story/show/180890663)

## Describe this PR

Add PaymentType property to WorkOrder and JobStatusUpdate entities.

Please also review the data migration to see if I havent considered anything.
- Any workOrder that isnt overtime will be set to 'bonus'. No workOrders will be 'CloseToBase' after the migration.
- There is a risk that there will be a delay between the migration and frontend sending the payment_type field in each request. If this does happen, I may need to manually run this migration again.
- Migration Down will update is_overtime to true if payment_type is overtime, otherwise false. This migration does assume that the is_archived field is fully up to date.

#### _Checklist_


RepairsApi/V2/Services/RepairsSnsService.cs
```diff
@@ -67,7 +73,7 @@ private async Task<List<EntityEventSns>> CreateMessages(WorkOrder workOrder, IEn
             {
                 var op = await _operativesGateway.GetByIdAsync(operative.OperativeId);
 
-                msgs.Add(new EntityEventSns
+                var snsMessage = new EntityEventSns
```
Asura5, line 76: Why is this necessary?

Asura5 (approved):

Asura5:
Ensure the changes to this and the listener are merged at the same time

RepairsApi/V2/Services/RepairsSnsService.cs
LBHCallumM: It isn't. I added a feature toggle, but later removed it.

teodor-hristoforov-hackney (approved):

PR #577, opened 2022-01-27
Hub VPC change for production

Asura5:
As previously advised by Matt K, we have changed the network setup within Hackney to route all traffic through the hub. While this change has already been completed, the TF definitions need to be updated as the VPC names have changed


LBHMKeyworth (approved):
Looks good to me.

PR #578, opened 2022-01-28
Lowercase user emails

Asura5:
- Lowercase email before database lookup + store in token
- E2E test (as the in-memory database used in unit tests is unreliable here)
- Data migration for existing records
- CHECK constraint to ensure all new emails are also lowercase


RepairsApi.Tests/E2ETests/OperativeE2ETests.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Threading.Tasks;
+using Newtonsoft.Json;
+using RepairsApi.V2.Boundary.Response;
+
+namespace RepairsApi.Tests.E2ETests
+{
+    public class OperativeE2ETests : OperativeWebApplicationFactory
+    {
+        [Test]
+        public async Task GetHubUser()
```
LBHCallumM, line 13: Shouldn't the test name specify that it's testing the capitalisation of the email address?

LBHCallumM (approved):

pixeltrix (approved):

PR #606, opened 2022-02-15
Filter planned priority jobs from entering DRS

teodor-hristoforov-hackney:
## Link to JIRA ticket

https://www.pivotaltracker.com/n/projects/2500129/stories/180645774

## Describe this PR

Filter planned priority jobs from entering DRS when creating new work orders.


RepairsApi/V2/Notifications/DRSNotificationHandler.cs
```diff
@@ -35,6 +35,10 @@ public class DRSNotificationHandler :
 
         private async Task<bool> UseDrs(WorkOrder workOder)
```
Asura5, line 36: Perhaps it is worth renaming this - you would expect a value of true returned from a method named "UseDrs" would indicate DRS is being used - whereas it's actually the opposite. So we could name it "SkipDrs" ?

teodor-hristoforov-hackney: I agree, it confused me as well

Asura5 (approved):

PR #616, opened 2022-02-22
Update DrsSync to Update Future WorkOrders by LastDrsSync

LBHCallumM:
## Describe this PR

Sometimes, not all future workOrders will be synced before the lambda function times out. 

This change will make sure that workOrders that are the most out-of-date will be updated first. This means, if some workOrders weren't updated in the previous sync, they will be the first in the next.

I have also renamed `UpdateHistoricalWorkOrders` to `UpdateFutureWorkOrders` because the name was incorrect.

#### _Checklist_


RepairsApi/V2/UseCase/SyncDRSWorkOrdersUseCase.cs
```diff
@@ -103,7 +103,8 @@ private async Task<IEnumerable<WorkOrder>> GetHistoricalWorkOrders()
             return workOrders
                 .Where(wo => wo.StatusCode != WorkStatusCode.Complete)
                 .Where(wo => !wo.HasManuallyAssignedOperatives())
-                .OrderBy(wo => wo.Appointments.Min(appt => appt.StartTime));
+                // oldest => newest. If null, it hasnt been synced yet, so it should be synced first
+                .OrderBy(wo => wo.LastDrsSync ?? DateTime.MinValue);
```
Asura5, line 107: Remember that only DLO workorders are linked to DRS. We have a class of workorders which will and should never be synced with DRS, so the above change contains a dangerous assumption - It will always prioritise the null ie..e non-DRS workorders for sync first, and if it is not successful the field may not be updated. This could seriously damage the performance of the sync system

PR #619, opened 2022-02-23
Move LegalDisrepair implementation to seperate controller method

LBHCallumM:
## Describe this PR

Performance issues were highlighted in the previous implementation. This is largely due to integrating with an external service (Google Sheets) of which we dont have control over.

To resolve this, I have created a separate endpoint to get LegalDisrepair information. This approach will allow the frontend to load this data in the background, reducing the tti.

New endpoint URL:
`properties/legalDisrepair/{propertyReference}`

Response object:
```
{
  propertyIsInLegalDisrepair: true
}
```

### Changes
- Move legalDisrepair logic to seperate endpoint
- Update tests
- Respond to previous PR comments

#### _Checklist_


teodor-hristoforov-hackney (approved):

RepairsApi/V2/Boundary/Response/LegalDisrepairResponse.cs
```diff
@@ -0,0 +1,12 @@
+namespace RepairsApi.V2.Boundary.Response
+{
+    public class LegalDisrepairResponse
+    {
+        public bool PropertyIsInLegalDisrepair { get; set; }
+
+        public LegalDisrepairResponse(bool propertyIsInLegalDisrepair)
+        {
+            PropertyIsInLegalDisrepair = propertyIsInLegalDisrepair;
```
Asura5, line 9: Don't want to include the property ID here? Perhaps a consideration for the future if we were to display the label on pages where there are multiple props

Asura5: Not essential for now though

RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -67,6 +68,27 @@ public async Task<IActionResult> GetProperty([FromRoute] GetPropertyRequest quer
...
 
+        /// <summary>
+        /// Retrieves a property
+        /// </summary>
+        /// <param name="query">The property reference</param>
+        /// <response code="200">Property found</response>
+        [HttpGet]
+        [Route("legalDisrepair/{propertyReference}")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(PropertyResponse), 200)]
```
Asura5, line 79: Produces 404 if the property can't be found?

RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -67,6 +68,27 @@ public async Task<IActionResult> GetProperty([FromRoute] GetPropertyRequest quer
...
+        [ProducesResponseType(typeof(string), StatusCodes.Status502BadGateway)]
```
Asura5, line 80: Produces 500 if there's issues

RepairsApi/V2/Services/GoogleSheets/GoogleSheetGateway.cs
```diff
@@ -22,13 +22,17 @@ public GoogleSheetGateway(SheetsService sheetsService, string spreadsheetId)
 
         public async Task<ValueRange> ReadRows(string range)
         {
+            Console.WriteLine($"GoogleSheetGateway ReadRows range[{range}]");
```
Asura5, line 25: Any reason this Console (which I assume was for local dev) and can't be replaced with logger?

RepairsApi/V2/Services/GoogleSheets/Mappers/LegalDisrepairData.cs
```diff
@@ -4,11 +4,11 @@ namespace RepairsApi.V2.Services.GoogleSheets.Mappers
 {
     public class LegalDisrepairData
     {
-        public List<string> PropertiesInLegalDisrepair { get; set; }
+        public HashSet<string> PropertiesInLegalDisrepair { get; set; }
```
Asura5, line 7: nice

RepairsApi/V2/Controllers/PropertiesController.cs
LBHCallumM: I could have, but I decided not to. Currently, it only returns whether a propertyReference is in the googleSheet. 

I could have first checked if the property exists, but that would then involve accessing the database. But I dont think this will ever occur because the properties page will return 404 before the api is called.

LBHCallumM: I could have, but I decided not to. Currently, it only returns whether a propertyReference is in the googleSheet. 

I could have first checked if the property exists, but that would then involve accessing the database. But I dont think this will ever occur because the properties page will return 404 before the api is called.

RepairsApi/V2/Services/GoogleSheets/GoogleSheetGateway.cs
LBHCallumM: I did this because of how I implemented GoogleSheetService. Since I am initilising the class in startup, I cannot easily add ILogger<T> into the constructor with dependency injection.

image

I could have solved this by using the ServiceProvider in the startup, and manually passing in the logger. However, I would need to do the same for GoogleSheetsGateway.

image

This was clearly a design flaw, but without redesigning the implementation of GoogleSheetsService, I thought `Console.WriteLine` was the simplest solution.

Alternatively, I could just remove the logging.

RepairsApi/V2/Controllers/PropertiesController.cs
LBHCallumM: I could have, but I decided not to. Currently, it only returns whether a propertyReference is in the googleSheet.

I could have first checked if the property exists, but that would then involve accessing the database. But I dont think this will ever occur because the properties page will return 404 before the api is called.

RepairsApi/V2/Controllers/PropertiesController.cs
LBHCallumM: I'm assuming you're not asking me to add StatusCodes.500 as a response, because no other endpoint does.

But I have removed the StatusCode.502 response because it doesn't call an external API, so it could only return 500.

PR #636, opened 2022-03-01
WorkOrder PaymentType should default to Bonus

LBHCallumM:
## Describe this PR

I noticed a small bug in the close-to-base change.

The issue is when creating a workOrder, paymentType defaults to null. This wont affect any business logic, and it wont affect the bonus calc.

But if you call `/workorders/1234` endpoint, the paymentType could be an empty string. I spoke to Neil, he doesn't think it will be an issue for the time being.

This change will ensure the paymentType is set to Bonus as default.

#### _Checklist_


NeilvB:
We should probably close this as external contractor work orders should not have a payment type. See https://www.pivotaltracker.com/n/projects/2500129/stories/181416900

PR #645, opened 2022-03-10
Move LocationAlerts and PersonAlerts to separate endpoints

LBHCallumM:
[Story](https://www.pivotaltracker.com/story/show/181511813)

## Describe this PR

The endpoint for getting LocationAlerts sometimes returns 504 BadGateway when it receives too many requests.

If this occurs, the properties endpoint will fail to load, preventing users from raising any repairs.

To resolve this issue, I have moved both alerts to separate endpoints. This means, if the API fails, users will still be able to raise a repair. I have separated them into two requests, rather than one, because personAlerts doesn't normally have issues, only the location alerts.

Added New Endpoints:

`/api/v2/properties/tenancyAgreementReference}/person-alerts`
`/api/v2/properties/{propertyReference}/location-alerts`

Response object for Both:
```
{
    "reference": "propRef or tenancyAgreementReferense",
    "alerts": [
        {
            "type": "",
            "comments": "",
            "startDate": "",
            "endDate": ""
        }
        // same as existing format
    ]
}
```

The `TenancyAgreementReference` value will be added to the property response `response.tenure.tenancyAgreementReference`.

#### _Checklist_


RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -68,6 +78,34 @@ public async Task<IActionResult> GetProperty([FromRoute] GetPropertyRequest quer
             return Ok(property.ToResponse());
         }
 
+        [HttpGet]
+        [Route("{tenancyAgreementReference}/person-alerts")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(AlertsViewModel), 200)]
```
Asura5, line 84: Can it also produce a 404 if the person isn't found, or a 400 if the request is malformed?

RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -68,6 +78,34 @@ public async Task<IActionResult> GetProperty([FromRoute] GetPropertyRequest quer
...
+        {
+            var alerts = await _personAlertsUseCase.ExecuteAsync(tenancyAgreementReference);
+
+            return Ok(alerts.ToResponse());
+        }
+
+        [HttpGet]
+        [Route("{propertyReference}/location-alerts")]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(AlertsViewModel), 200)]
```
Asura5, line 98: Can it also produce a 404 if the location isn't found, or a 400 if the request is malformed?

RepairsApi/V2/Factories/ResponseFactory.cs
```diff
@@ -31,6 +31,24 @@ public static CautionaryAlertResponseList ToResponse(this AlertList domain)
             };
         }
 
+        public static AlertsViewModel ToResponse(this PropertyAlertList domain)
```
Asura5, line 34: SOLID - breaks DRY. Refactor available: have a common superclass for PropertyAlertList & PersonAlertList and use `typeof` to set the refernces

RepairsApi/V2/Controllers/PropertiesController.cs
LBHCallumM: You're right about the 404 response. But neither endpoint has any validation for the propertyReference or tenancyAgreementReference.

RepairsApi/V2/Factories/ResponseFactory.cs
LBHCallumM: Also, the ListAlerts endpoint isn't used anymore. I checked cloudwatch for any logs within the last 2 months.

RepairsApi/V2/Controllers/PropertiesController.cs
LBHCallumM: Discussed not to include 404.

LBHCallumM: Added whitespace checks which will returns badRequest.

pixeltrix (approved):

PR #650, opened 2022-03-17
Small PR to move reused code into WebApplicationFactory class

LBHCallumM:
## Describe this PR

Small PR to move reused code into WebApplicationFactory class


RepairsApi.Tests/MockWebApplicationFactory.cs
```diff
@@ -60,15 +61,27 @@ public class MockWebApplicationFactory
...
         public MockWebApplicationFactory()
             : base()
         {
             UserGroup = UserGroups.Agent;
+            ConfigureFixture();
+        }
+
+        private void ConfigureFixture()
+        {
+            _fixture.Behaviors.OfType<ThrowingRecursionBehavior>().ToList()
```
Asura5, line 82: This isn't quite doing the same thing. This is removing all Behaviors of the type ThrowingRecursionBehavior, whereas originally a single ThrowingRecursionBehavior is removed. Have you checked that this is a set and therefore the behaviour (lol) is the same?

PR #673, opened 2022-03-30
Add option to get all SorCodes using a filter in ListSorCodes endpoint

LBHCallumM:
## Link to JIRA ticket

[Story](https://www.pivotaltracker.com/story/show/181508969)

## Describe this PR

Update endpoint ListSorCodes to enable listing sorCodes within all trades, but using a filter.

Url: `/api/v2/schedule-of-rates/codes`

QueryParameters:
```
{
  "showAllTrades": false,
  "filter": "",
  ... same as existing object
}
```

Important changes:
- The parameter `showAllTrades` will override `showAdditionalTrades`
- When `showAllTrades` is `true`, `filter` will be validated. It must be at least 3 characters, otherwise 400BadRequest will be returned.

As discussed with Neil, the frontend uses its own sort/filter mechanism. This means, the sorCodes will only be loaded once with the first three characters. Any more characters will be filtered through the frontend. So there wont be a request for every letter typed by the user.

Implementation:
So that I can breakup the query, I have created the class `ScheduleOfRatesDTO`, which will replace the anonymous object `new { sor, sorContract };`. I am not sure where the best place is for this class.


RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
```diff
@@ -61,53 +61,78 @@ select trade
...
                 },
-                Cost = x.sorContract.Cost ?? x.sor.Cost
+                Cost = x.SorContract.Cost ?? x.SOR.Cost
             }).ToListAsync();
         }
 
+        private static IQueryable<ScheduleOfRatesDTO> FilterByAllTrades(IQueryable<ScheduleOfRatesDTO> baseQuery, string filter)
+        {
+            return baseQuery
+                .Where(x => x.SOR.Code.Contains(filter) || x.SOR.ShortDescription.Contains(filter));
```
Asura5, line 115: Is adding the ShortDescription filter a risk here? It's not functionality that was present before. Is there could be a random work in the string that contains a hit for the three letter code? Or is that the whole point?

RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
```diff
@@ -61,53 +61,78 @@ select trade
...
+        }
+
+        private async Task<IQueryable<ScheduleOfRatesDTO>> FilterByAdditionalTrades(IQueryable<ScheduleOfRatesDTO> baseQuery, string tradeCode)
+        {
+            var trade = await _context.Trades.FindAsync(tradeCode);
+
+            var relatedTrades = trade.RelatedTrades.Select(x => x.SecondaryTradeId);
+
+            baseQuery = baseQuery.Where(
```
Asura5, line 124: There does seem to be an inconsistency here - FilterAllTrades filters by code, description but FilterAdditionalTrades just filters by code?

RepairsApi.Tests/V2/Controllers/ScheduleOfRatesControllerTests.cs
```diff
@@ -60,6 +60,32 @@ public async Task ListSorCodes()
             response.Should().BeEquivalentTo(model);
         }
 
+        [TestCase(null, 400)]
+        [TestCase("", 400)]
+        [TestCase("   ", 400)]
+        [TestCase("a", 400)]
+        [TestCase("aaa", 200)]
+        [TestCase("aaaaa", 200)]
+        public async Task ListSorCodes_WhenShowAllTradesTrue_ValidatesFilter(string filter, int statusCode)
```
Asura5, line 69: In teh PR description, you state that "The parameter showAllTrades will override showAdditionalTrades" is an important change. If that's the case, might it be worth adding a test for this?

Asura5 (approved):

PR #679, opened 2022-04-05
Endpoint for 'Upload multiple SOR codes'

LBHCallumM:
## Link to JIRA ticket
[Story](https://www.pivotaltracker.com/story/show/181730893)

## Describe this PR

Add a new endpoint to allow a user to upload multiple SorCodes.

After being parsed by the frontend, a list of sorCodes will be sent to this endpoint. The endpoint will return any valid sorCodes with the full dataset. The response is identical to the endpoint `/codes`.

### CheckSorCodes
URL: `GET /api/v2/schedule-of-rates/check`

Parameters:
- `propertyReference` [required]
- `contractorReference` [required]
- `sorCode` [required] [accepts multiple e.g. `/?sorcode=1234&sorCode=abcd`]
- `isRaisable`

The endpoint will return 200 if no sorCodes are in the request.

If no sorCodes are sent in the request, bad request is returned.

Initially I wasnt going to include contractorReference or propertyReference in the request, but I decided it makes more sense to use the same filters as when getting sorCodes directly.


RepairsApi/V2/Controllers/ScheduleOfRatesController.cs
```diff
@@ -118,5 +119,29 @@ public async Task<IActionResult> ListPriorities()
...
         }
+
+
+        /// <summary>
+        /// Returns list of SOR codes
+        /// </summary>
+        /// <response code="200">Success. Returns a list of SOR codes</response>
+        /// <response code="400">Invalid Query Parameter.</response>
+        [ProducesResponseType(typeof(IEnumerable<ScheduleOfRatesModel>), StatusCodes.Status200OK)]
+        [Route("populate")]
```
Asura5, line 130: I would agree it's not really populate - it's more of a /check

Asura5 (approved):
Looks good except the route name

PR #681, opened 2022-04-06
Add check for manual operative assignment after individual GET

Asura5:


RepairsApi/V2/UseCase/SyncDRSWorkOrdersUseCase.cs
```diff
@@ -148,6 +148,12 @@ public async Task<bool> ValidateWorkOrder(WorkOrder wo)
         {
             const int PlannedPriorityCode = 9;
 
+            if (wo.HasManuallyAssignedOperatives())
```
LBHCallumM, line 151: Pretty sure this line is being called inside `GetWOrkOrdersFromToday` and `GetHistoricalWorkOrders`.

LBHCallumM (approved):
Approved, except for the other comment

RepairsApi/V2/UseCase/SyncDRSWorkOrdersUseCase.cs
LBHCallumM: sure

Asura5: join drop-in

PR #687, opened 2022-04-11
DRS Sync - more tests + fewer exceptions

Asura5:
No additions to functionality, but in order to help diagnose problems in the future, this handles the First() exception which clutters the exception log. Also, add explicit tests for two scenarios as yet not covered:

Scenario 1: 1 booking, in the future => sync the single item in the future
Scenario 2: 1 booking, today => sync the single item today
**Scenario 3**: 2+ bookings, 1 booking today and 1 in the future (multi-day job for single operative) => sync today's item
**Scenario 4**: 2 bookings, both today (single day, multi-operative job) => sync both items today
Scenario 5: 3+ booking, 2 today (multi-day job for multiple operatives) => sync both items today


RepairsApi/V2/Services/DRS/DrsService.cs
```diff
@@ -221,12 +221,14 @@ public async Task UpdateWorkOrderDetails(int workOrderId)
 
             var theBooking = ExtractBooking(order);
 
-            var theResources = ExtractBookingResources(order);
+            if (theBooking == null) return;
 
-            _logger.LogInformation($"UpdateWorkOrderDetails found {theResources.Count()} resources to update for workOrderId: {workOrderId}");
+            var theResources = ExtractBookingResources(order);
 
             if (theResources.IsNullOrEmpty()) return;
```
LBHCallumM, line 228: Would there be any benefit to logging here if no resources are found?

Asura5: Maybe, maybe not, so I've added it anyway

PR #695, opened 2022-04-20
Update both (all) DRS appointments when a multi-operative job is closed

Asura5:
- Existing tests automatically generate 2+ bookings (via hand rolled ObjectFaker class) so all tests are automatically already set up for multiple bookings. Existing tests have been modified to check subsequent bookings


RepairsApi/V2/Services/DRS/IDrsMapping.cs
```diff
@@ -10,6 +11,7 @@ public interface IDrsMapping
     {
         Task<createOrder> BuildCreateOrderRequest(string sessionId, WorkOrder workOrder);
         Task<deleteOrder> BuildDeleteOrderRequest(string sessionId, WorkOrder workOrder);
-        Task<updateBooking> BuildCompleteOrderUpdateBookingRequest(string sessionId, WorkOrder workOrder, order drsOrder, string bookingStatus = DrsBookingStatusCodes.Completed);
+        Task<updateBooking> BuildCompleteOrderUpdateForSingleBooking(string sessionId, WorkOrder workOrder, order drsOrder, booking drsBooking, string bookingStatus = DrsBookingStatusCodes.Completed);
```
LBHCallumM, line 14: Will the method `BuildCompleteOrderUpdateForSingleBooking` ever be called directly, other than when being called by `BuildCompleteOrderUpdate`? I don't think its needed in the interface.

LBHCallumM (approved):

RepairsApi/V2/Services/DRS/IDrsMapping.cs
Asura5: Updated

PR #699, opened 2022-04-26
Add DLO BudgetCodes + Update GetBudgetCodes endpoint

LBHCallumM:
## Link to JIRA ticket

[Story](https://www.pivotaltracker.com/story/show/181446341)

### Update GetBudgetCodes endpont

URL: `/api/v2/workOrders/budget-codes`

Endpoint now **doesn't** require the parameter `contractorReference`, previously this was mandatory.

If `contractorReference` is null, all budgetCodes will be returned.

### Migration - Add DLO BudgetCodes

[Spreadsheet](https://docs.google.com/spreadsheets/d/1IJ7JVZ7MLNE_MKn-K4q6x18pfS9VeODtfh_CddaSWYw/edit#gid=1198572254) (Tab = First release DLO)

Import remaining budgetCodes, and assign them to the associated contractor.

I had some difficulty getting the id from each newly inserted budgetCode, and using it when associating a contractor. This is why I have resulted to using raw SQL with forloops. If you can think of a better solution, I'm all ears.


RepairsApi.Tests/E2ETests/GetBudgetCodesE2ETests.cs
```diff
@@ -36,17 +36,10 @@ public void SetUp()
...
         {
-            // Arrange
-            var contractorId = string.Empty;
-
-            // Act
-            var response = GetBudgetCodesRequest(contractorId);
-
-            // Assert
-            response.Status.Should().Be(HttpStatusCode.BadRequest);
+            DeleteAllBudgetCodes();
```
pixeltrix, line 42: Is this necessary? Usually the test is run inside a transaction and then the rollback is issued at the end of the test. However RH uses the InMemory database provider for E2E tests rather than PostgreSQL so the behaviour may be different.

pixeltrix (approved):
Other than the query about having to manually remove database objects looks good 👍🏻

RepairsApi.Tests/E2ETests/GetBudgetCodesE2ETests.cs
LBHCallumM: Yes it should, but the E2E tests were failing when I ran them together.

PR #708, opened 2022-05-09
Callum/tidy up/cleanup workorder entity

LBHCallumM:
## Describe this PR

Many columns in the WorkOrder entity are redundent, so I have created this PR to remove them, and any associated objects/factories.

Here is a list of all the columns that will be removed in this PR. I have checked each on prod, they are all null, except for "work_class_work_class_code" which is `0`:

- "access_information_description"
- "access_information_keysafe_code"
- "access_information_keysafe_location"
- "date_reported"
- "estimated_labor_hours"
- "location_of_repair"
- "parking_arrangements"
- "work_class_work_class_code"
- "work_class_work_class_description",
- "work_class_work_class_sub_type_work_class_sub_type_description"
- "work_class_work_class_sub_type_work_class_sub_type_name"
- "work_type"


Asura5:
What is your basis for stating that these columns are unused?

LBHCallumM:
> What is your basis for stating that these columns are unused?

I selected these columns for two reasons:
1. In the production database, they only contain null or default values. So clearly they aren't being used to display information to the user.
2. The properties in the workOrder entity are only referenced in factory method, no logic.

LBHCallumM:
> And you are confident that EF is not going to complain about the removal?

Well, the migration only involves dropping a few columns. Is there reason that I should be concerned about this not working?

Also, is there a way to test this without merging this to dev?

PR #713, opened 2022-05-16
Customize AutoFixture to speed up tests

LBHCallumM:
## Describe this PR

The tests in RH have started to take longer over time. This was because of the complex relationships between these objects, causing a lot of unnecessary work for for AutoFixture.

I found a solution on stackoverflow that will not include any properties with the `virtual` attribute. This can be added when initilising AutoFixture within a test class.

```
 _fixture.Customizations.Add(new IgnoreVirtualMembersSpecimenBuilder());
```


RepairsApi.Tests/V2/UseCase/UpdateJobStatusUseCaseTests.cs
```diff
@@ -92,15 +98,26 @@ public void DoesNotThrowUnsupportedWhenOtherTypeCode()
...
+            _jobStatusUpdateGateway
+                .Setup(gateway => gateway.CreateJobStatusUpdate(It.Is<JobStatusUpdate>(jsu => jsu.RelatedWorkOrder.Id == desiredWorkOrderId)))
                 .ReturnsAsync(1);
 
             var workOrder = _fixture.Create<WorkOrder>();
+
             workOrder.Id = desiredWorkOrderId;
 
-            _repairsGatewayMock.Setup(gateway => gateway.GetWorkOrder(It.Is<int>(i => i == desiredWorkOrderId)))
+            workOrder.WorkElements = new List<WorkElement>
```
Asura5, line 111: This test looks like it has changed implementation due to the customisation - is that the case?

LBHCallumM: You are correct. The list of workElements wont be loaded by default.

This can either be manually added in the test, or ignore the the customization for the test class.

PR #723, opened 2022-05-24
Import Voids SorCodes for DLO

LBHCallumM:


pixeltrix:
If you change the `V1` to `V2` that should fix the coverage failure from SonarCloud

https://github.com/LBHackney-IT/repairs-api-dotnet/blob/7f46c39db87c2dc1854df0e5335a148b396d30fb/RepairsApi/RepairsApi.csproj#L60-L65

LBHCallumM:
> If you change the `V1` to `V2` that should fix the coverage failure from SonarCloud
> 
> https://github.com/LBHackney-IT/repairs-api-dotnet/blob/7f46c39db87c2dc1854df0e5335a148b396d30fb/RepairsApi/RepairsApi.csproj#L60-L65

Thanks, that worked.

Asura5 (approved):

PR #728, opened 2022-05-30
Fix migration - remove duplicate SorCodes 

LBHCallumM:
I didn't spot a couple of duplicate SorCodes, which caused a foreign key constraint violation


pixeltrix (commented):
I don't see either of those codes in the production database and you appear to have removed both lines? Shouldn't one of them be kept?

LBHCallumM:
> I don't see either of those codes in the production database and you appear to have removed both lines? Shouldn't one of them be kept?

Yes, but I don't know which ones to add. It will be easier to just add them later.

pixeltrix (approved):

PR #736, opened 2022-06-01
Dave/move req completion dt backend

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://www.pivotaltracker.com/n/projects/2500129/stories/182232706)

### Endpoint ScheduleRepair

URL: POST `/api/v2/workOrders/schedule` || `/api/v2/repairs/schedule`

#### Request Object

The request object is identical, however, some of the validation has changed.

Either:
1. Pass a value in the field `request.priority.requiredCompletionDateTime`. This will be used for the workOrder.
2. Pass a value in the field `request.dateRaised`, and the requiredCompletionDateTime will be calculated on the backend.

#### Other changes

I have moved the logic from `DBModelFactory`  to `CreaseWorkOrderUseCase` because the logic became too complicated.

Also, the ACs specify that workOrders raised on a Saturday should be recorded as if they had been raised on the previous Friday. However, with the implementation of the working days feature, this is irrelevant because a workOrder raised on Friday or Saturday will result in the same target date.

Lastly, I have included the voids priorities, which weren't included in the ACs.


RepairsApi/V2/Gateways/BankHolidayGateway.cs
```diff
@@ -0,0 +1,38 @@
...
+        private readonly IApiGateway _apiGateway;
+        private readonly ILogger<BankHolidayGateway> _logger;
+
+        public BankHolidayGateway(IApiGateway apiGateway, ILogger<BankHolidayGateway> logger)
+        {
+            _apiGateway = apiGateway;
+            _logger = logger;
+        }
+
+        public async Task<HashSet<DateTime>> GetBankHolidays()
```
Asura5, line 23: These are not going to change very often - in fact very rarely. Consider adding a [response cache](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.responsecacheattribute) to increase the performance dramatically

Asura5 (approved):

PR #738, opened 2022-06-07
Change order of Priority Codes DropDown

LBHCallumM:
## Link to JIRA ticket

[Story](https://www.pivotaltracker.com/story/show/182302282)

## Describe this PR

PriorityCodes in the dropdown should be displayed in the following order:

1. [I] IMMEDIATE
2. [E] EMERGENCY
3. [U] URGENT
4. [N] NORMAL
5. [N60] SURVEYING ONLY
6. [P] PLANNED MAINT
7. [V15] VOIDS MINOR
8. [V30] VOIDS MAJOR

The existing implementation orders these by `daysToComplete`. 

My solution is to add a new column `displayOrder`, and update the 8 main priority codes in that order.


RepairsApi.Tests/V2/Gateways/SorPriorityGatewayTests.cs
```diff
@@ -18,36 +23,93 @@ public void Setup()
...
+            // arrange
+            var numberOfPriorities = 10;
+            var priorities = GeneratePrioritiesWithRandomDisplayOrder(numberOfPriorities);
+            await SavePrioritiesToDb(priorities);
 
             // act
             var codes = await _classUnderTest.GetPriorities();
 
             // assert
             codes.Should().NotBeNullOrEmpty();
```
Asura5, line 55: This can be removed as it's implicitly checked by the other Asserts

PR #743, opened 2022-06-15
Allow null contractor to raise repair

elena-evans:
[## Link to JIRA ticket

Add a link to the JIRA ticket that the changes in this PR describe.](https://www.pivotaltracker.com/story/show/182463380)

## Describe this PR

Repairs Online will not pass a contractor reference from the front end (previously mandatory in Repairs Hub to find the contract reference).

Change the endpoint so that it finds an applicable contract reference if none is supplied
Added a test to ensure SOR codes are still found when no contractor reference is supplied.


RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
```diff
@@ -125,16 +125,21 @@ public async Task<IEnumerable<ScheduleOfRatesModel>> CheckSorCodes(CheckSorCodes
...
+                            where
+                                contract.PropertyMap.Any(pm => pm.PropRef == propertyReference) &&
+                                contract.EffectiveDate < DateTime.UtcNow && DateTime.UtcNow < contract.TerminationDate
+                            select new { SOR = sor, SorContract = sorContract, ContractorReference = contractorReference };
+
+            if (contractorReference != null)
+            {
+                baseQuery = baseQuery.Where(x => x.ContractorReference == contractorReference);
+            }
+            return baseQuery.Select(x => new ScheduleOfRatesDTO { SOR = x.SOR, SorContract = x.SorContract });
```
LBHCallumM, line 142: Do you think this will return a lot of SorCodes? It could be a performance issue. It might be worth making sure the incremental search is used on the frontend when the contractorReference is null.

RepairsApi.Tests/V2/Gateways/ScheduleOfRatesGatewayTests.cs
```diff
@@ -439,6 +439,27 @@ await SeedTrade("CCC")
             response.Should().Contain(x => x.Code == tradeCodes[4].Code);
         }
 
+        [Test]
+        public async Task GetSorCodes_WhenContractorIsUnknown_ReturnsDefaultContractIfOneExists()
```
LBHCallumM, line 443: This change also affects the `CheckSorCodes` method. You should also add tests for that.

RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
elena-evans: The usual workflow should not have a null contractorReference. This should only be the case when raising a repair using Repairs Online, in which case the SOR code will first be selected via a series of questions and then used to find a relevant contract

LBHCallumM (approved):

PR #754, opened 2022-06-21
Prevent OOH Gas WorkOrders from going to DRS

LBHCallumM:
## Link to JIRA ticket

[Story](https://www.pivotaltracker.com/story/show/182387829)

## Describe this PR

Add a check to prevent workOrders from going to DRS when:

- The trade is OOH
- The contractor is "HH Gas Breakdown - H04"


pixeltrix (approved):
This is fine for now but I wonder whether adding a flag to the contracts table for whether work orders should be scheduled via DRS would be better in the long term.

LBHCallumM:
> This is fine for now but I wonder whether adding a flag to the contracts table for whether work orders should be scheduled via DRS would be better in the long term.

Thanks for the suggestion. I will look into it

PR #757, opened 2022-06-23
Fix InvalidOperationException when more or less than one item

thecallum:
## Describe this PR

The method `.Single()` throws an exception if an array doesn't contain exactly 1 item. I have replaced it with `FirstOrDefault()`.


thecallum:
Wrong account

PR #765, opened 2022-06-28
Temporarily add housing search into existing controller

Asura5:
- The idea here is that we can do the same search, but just add `useNewHousingSearch=true` to use the HS functionality. This will allow more comprehensive testing with end users


LBHCallumM (approved):

Asura5:
Waiting for Availability by Trade to be released

PR #778, opened 2022-07-05
Modify last appointment if 2+ appts

Asura5:
- For some reason we are getting duplicate appointments in the DB for the same workorder.
- This has happened very occasionally (4 times in 1 year)
- So, use the latest workorder by default


RepairsApi/V2/Gateways/AppointmentGateway.cs
```diff
@@ -75,7 +75,7 @@ public async Task SetTimedBooking(int workOrderId, DateTime startTime, DateTime
                 throw new ResourceNotFoundException(Resources.WorkOrderNotFound);
             }
 
-            var existingAppointment = await _repairsContext.Appointments.SingleOrDefaultAsync(a => a.WorkOrderId == workOrderId);
+            var existingAppointment = await _repairsContext.Appointments.LastOrDefaultAsync(a => a.WorkOrderId == workOrderId);
```
LBHCallumM, line 78: Should we still log a message if there are multiple appointments? It could be useful for debugging.

LBHCallumM (approved):

PR #803, opened 2022-07-26
Add HousingSearchMatchType to new search endpoint

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/MR-397?atlOrigin=eyJpIjoiMTQwMzllMjc0YThjNGY5NWExMGI5ODY1ZjcwYTJiNGEiLCJwIjoiaiJ9)

## Describe this PR

To re-release the search, we need to implement pagination of the results (Sometimes more than 50 addresses are returned for a postcode). This was previously implemented on the Frontend with a Feature Toggle.

The frontend FT will call `/properties/search` instead of `/properties` when the FT is enabled. Therefore, I have added `housingSearchMatchType` to the new endpoint.


Asura5 (approved):
You'll need a separate PR to send the new search type

LBHCallumM:
> You'll need a separate PR to send the new search type

Sure. What value should I be sending?

PR #819, opened 2022-08-02
Adjust DrsSync cron schedule to 6am-8pm

LBHCallumM:
## Describe this PR

Cyndi requested for the DRS sync cron schedule to be updated to 6am-8pm


RepairsApi/serverless.yml
```diff
@@ -39,7 +39,7 @@ functions:
       DrsOptions__Login: ${ssm:/${self:service}/${self:provider.stage}/drs-login}
       DrsOptions__Password: ${ssm:/${self:service}/${self:provider.stage}/drs-password}
     events:
-      - schedule: cron(0/20 5-19 ? * * *)
+      - schedule: cron(0/20 6-20 ? * * *)
```
pixeltrix, line 42: Is this in UTC? If so then `5-19` is already 6am to 8pm in BST

PR #821, opened 2022-08-03
Callum/adjust cron schedule 3

LBHCallumM:
## Describe this PR

DRS Sync should run 06:00 - 20:00 during BST or GMT.


pixeltrix (approved):

pixeltrix:
I'd double check this in AWS after it's been applied

PR #825, opened 2022-08-09
Update target date calculation

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/MR-507?atlOrigin=eyJpIjoiOGFlNjQwZGE1M2FkNGQ5ZmIzNWNlNmZhYzk0M2FiYjciLCJwIjoiaiJ9)

## Describe this PR

The previous calculation for target dates was incorrect.

If a job is raised on a non-working day (weekend or bank holiday), the target date should include an additional day, and start at 8am.

For example:

A repair is raised on Saturday 1st January at 10pm. The priority is Normal, which is 5 working days.

5 working days would result in a target date of Friday 7th January. However, this doesn't account for the fact that the work cannot be completed on that Saturday.


Asura5 (approved):
Only suggestion would be to abstract to helper

PR #826, opened 2022-08-10
Update Location Alerts Integration

LBHCallumM:
## Describe this PR

There were a couple of changes that needed to be made the the location alerts (cautionary alerts) integration.

- The API doesnt return 404 when no alerts are found anymore, so this check was removed
- The URL isnt going to change anymore, so the gatewayOption `AlertsApiPropertyAlertsEndpoint` can be removed
- Updated the logging to use the correct template format


RepairsApi.Tests/E2ETests/LocationAlertsE2ETests.cs
```diff
@@ -45,23 +45,5 @@ public void LocationAlerts_WhenValid_ReturnsAlerts()
             response.Content.Reference.Should().Be(propertyReference);
             response.Content.Alerts.Should().HaveCount(numberOfAlerts);
         }
-
-        [Test]
-        public void LocationAlerts_WhenEmptyResponse_DoesntThrowException()
```
Asura5, line 50: Is there a way of fixing them instead of removing them?

LBHCallumM: You're right. I have re-added and fixed the test.

Asura5 (approved):

PR #832, opened 2022-08-23
Return Contractor List for Budget Codes

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/MR-410?atlOrigin=eyJpIjoiYmM1OWFmYzYzOGM2NGM4ODhjZWYyZWM2ZTQxODFkYTgiLCJwIjoiaiJ9)

## Describe this PR

Add `ContractorList` to budgetCode response when returning all budgetCodes (when contractorReference not specified in the request). If the contractor has been specified, such as Purdy, we dont need to display it in the dropdown list because it is implicit.

When looking at the code, I discovered that the logic exists to return budgetCodes mapped to each contractor. However, I decided against this when I read an old comment From Raffaella.

'_We need to display ALL the budget codes presented on the tab "First release DLO" [link] without validating the contractor. The Budget code feature is enabled only to a super users group who know the codes and which one much be selected for a specific property/trade/contractor_'

By displaying the additional information, we will meet the requirement by allowing the user to differentiate between the 'identical' budget code options.

Secondly, since the contractor data is already in the database (budget_codes_contractor I believe), I was able to map them and include them in the response as a string. Hence no migration.

## Response Object

Route: [GET] `/api/workOrders/budget-codes/`
```
[
  {
      ...,
      "contractorsList": "H01, H02"
  }
]
```


RepairsApi/V2/Gateways/BudgetCodeGateway.cs
```diff
@@ -34,7 +34,19 @@ public async Task<IEnumerable<Domain.BudgetCode>> GetBudgetCodes(GetBudgetCodesR
...
+                .Include(x => x.Contractors)
+                .Select(q => new BudgetCodeWithContractorList
+                {
+                    Id = q.Id,
+                    ExternalCostCode = q.ExternalCostCode,
+                    CorporateSubjectiveCode = q.CorporateSubjectiveCode,
+                    DescriptionOfWorks = q.DescriptionOfWorks,
+                    Contractors = q.Contractors,
+                    CostCode = q.CostCode,
+                    ContractorsList = string.Join(", ", q.Contractors.Select(x => x.Reference))
```
Asura5, line 47: Is there an assumption here that every budget code will have at least one contractor?

LBHCallumM: Actually, Ignore what I said. The test ` GetBudgetCodes_WhenContractorReferenceNullAndMultipleContractors_ReturnsContractorListString` tests when there are no contractors assigned to a budget code. It returns an empty string

Asura5 (approved):

PR #833, opened 2022-08-24
Re-add old search

LBHCallumM:


RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -165,5 +168,37 @@ public async Task<IActionResult> Search([FromQuery] SearchForPropertiesRequest r
...
+        /// <param name="postcode">A postcode</param>
+        /// <param name="q">A postcode or partial or full address</param>
+        /// <response code="200">Properties found</response>
+        [HttpGet]
+        [Produces("application/json")]
+        [ProducesResponseType(typeof(List<PropertyListItem>), 200)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status502BadGateway)]
+        [ProducesDefaultResponseType]
+        [Authorize(Roles = UserGroups.Agent + "," + UserGroups.ContractManager + "," + UserGroups.AuthorisationManager + "," + UserGroups.Operative)]
+        public async Task<IActionResult> ListProperties([FromQuery] string address, [FromQuery] string postcode, [FromQuery] string q)
```
Asura5, line 187: How will this is activated by the frontend. If this is implemented as shown, is the controller going to choose this route over the new HS API route in the same controller? Are they runnable side by side or will you have to switch a config item on and off to change the behaviour?

LBHCallumM: There is a FT on the frontend, which updates which search endpoint to use

Asura5: OK - please explain to the UX/UR team in slack what this means in practical terms for them, if they are not going to be able to switch old/new search without developer involvement

Asura5 (approved):

PR #835, opened 2022-08-25
Update search to include PostPreamble in the address field

LBHCallumM:
Update search to include PostPreamble in the address field


RepairsApi/V2/Factories/ResponseFactory.cs
```diff
@@ -404,15 +404,30 @@ public static SearchForPropertiesResponse ToResponse(this HousingSearchAPIRespon
...
             {
-                Address = asset.AssetAddress.AddressLine1,
+                Address = address,
                 PostCode = asset.AssetAddress.PostCode,
                 PropertyReference = asset.AssetId,
                 PropertyType = asset.AssetType.ToString()
             };
         }
 
+        private static string FormatAssetAddress(Asset asset)
```
Asura5, line 418: Have you checked that any properties have the same thing in PostPreable and addressline1 as I think that's why it was excluded in the first place

LBHCallumM: This is how its displayed in MMH search, so I dont think so. Also, I don't how how I would check for any of these properties.

Asura5: "I don't know how to do it so I won't do it" ?

Asura5 (approved):

PR #840, opened 2022-09-07
Enable strict search filtering

Asura5:
- Only include a search result if it contains the search term
- Adapted from https://github.com/LBHackney-IT/property-information-api/blob/9821caba6e158dda898a31cffc8710e94d234974/PropertyInformationApi/V1/Gateways/PropertyGateway.cs


LBHCallumM (approved):
Looks good. Is there anything else that can be done to normalise the input format? For example, ensuring the postcode contains a space.

Asura5:
> Looks good. Is there anything else that can be done to normalise the input format? For example, ensuring the postcode contains a space.

Or removing all spaces?

PR #845, opened 2022-09-12
Add additional codes for Leaseholder project and uplift DLO codes

pixeltrix:


pixeltrix:
#843 will probably need merging before this one

LBHCallumM (approved):
I've had a skim through, but if there is a typo, I will most likely have missed it.

You should try running the migration against the database `repairs-db-db-local`. Its a clone of the dev database, and has the same connection string other than the name.

pixeltrix:
> You should try running the migration against the database `repairs-db-db-local`. Its a clone of the dev database, and has the same connection string other than the name.

I dumped the schema from production and the contractors, contracts, sor_codes & sor_contracts tables - seems to run fine and I've tried to make the queries idempotent wherever possible.

Asura5 (approved):

pixeltrix:
Updated with:
- additional Axis codes for drains, estate roads and paths enabled
- links inserted to the new contracts in the `property_contracts` table

PR #850, opened 2022-09-16
Enable a pound code for Pest Control

pixeltrix:
Whilst more specific codes are agreed upon allow Pest Control to continue with current practice.


pixeltrix:
#845 should be merged before this one but deployed together to production as it terminates the `001-ENV-EEHS` contract in favour of the new `F22-ENV-EEHS` contract.

Asura5 (approved):
I'm really not sure how to review these massive SQL changes other than to say - it runs fine against local

PR #861, opened 2022-09-22
Check for null PostPreamble in search filtering

LBHCallumM:


RepairsApi.Tests/V2/Helpers/AddressComparerTests.cs
```diff
@@ -0,0 +1,43 @@
...
+        public void Compare_WhenPostPreambleNull_DoesntThrowException()
+        {
+            // Arrange
+            var address1 = new AssetAddress
+            {
+                AddressLine1 = "218 Lea View House  Springfield",
+                PostPreamble = null
+            };
+
+            var address2 = new AssetAddress
```
Asura5, line 30: Are these intended to be the same?

Asura5 (approved):

RepairsApi.Tests/V2/Helpers/AddressComparerTests.cs
LBHCallumM: Yes. The issue occurs when one or both of the postPreambes are null. They are checked when addressLine1 matches.

PR #871, opened 2022-09-28
Normalize postcode in SearchForPropertiesModel

LBHCallumM:
## Describe this PR

If a user enters a postcode into the search, but it doesn't match the postcode in ElasticSearch, they wont get any results.

To resolve this, I have added the following logic to `SearchForPropertiesModel.GetQueryParameters()`:
 1. Check if the search query is **only a postcode**. I won't be able to normalize the search query if the user searches '53 Pitcairn House E9 6PU`. That is too specific.
 2. If the search query is a postcode, I can normalise it to the correct format.

The logic should only run if the search query matches a regex condition. If it doesn't match, it should run as normal.


RepairsApi/V2/Helpers/PostCodeHelpers.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public static class PostCodeHelpers
+    {
+        public static string NormalizePostcode(string postcode)
+        {
+            //removes end and start spaces 
+            postcode = postcode.Trim();
+
+            //removes in middle spaces 
+            postcode = postcode.Replace(" ", "");
```
Asura5, line 13: Isn't this going to also do the work of the Trim() ?

RepairsApi/V2/Helpers/PostCodeHelpers.cs
```diff
@@ -0,0 +1,36 @@
...
+                5 => postcode.Insert(2, " "),
+                6 => postcode.Insert(3, " "),
+                7 => postcode.Insert(4, " "),
+                _ => postcode,
+            };
+        }
+
+        public static bool SearchTextIsValidPostCode(string searchText)
+        {
+            var trimmed = searchText.Trim();
```
Asura5, line 28: If length > 7 immediately validate as not a postcode?

RepairsApi/V2/Helpers/PostCodeHelpers.cs
LBHCallumM: Yes, you're right

RepairsApi/V2/Helpers/PostCodeHelpers.cs
LBHCallumM: The TDD in me wants to say no because it passes the tests, but I can add it anyway.

Asura5 (approved):

PR #875, opened 2022-10-05
Fix gas_service_data mapping

LBHCallumM:
## Describe this PR

The `GasServiceData` table was not correctly mapped.

image


Asura5 (approved):
Here's me thinking it was pluralised for EF

PR #876, opened 2022-10-05
Remove SoR contracts for unused H01 codes

pixeltrix:
The HO1 Reactive Repairs team have identified the list of codes they require so remove any that aren't in that list. The down migration just replays the SoR contracts uplift insert statment and relies on the 'ON CONFLICT' clause to only restore the missing codes.


Asura5 (approved):
£10 says we'll see a 'code missing' Halo ticket within a fortnight

PR #879, opened 2022-10-07
Callum/gas servicing create or update

LBHCallumM:
## Describe this PR

I decided to implement separate routes for the create and update methods because it breaks the Single Responsibility Principle.

## Create
`POST /api/v2/gasservicedata` returns:

```cs
public class SaveGasServiceDataResponse
  {
      public int Id { get; set; }
      public int Version { get; set; }
  }
```

## Update
`POST /api/v2/gasservicedata/{id}` returns:

```cs
public class SaveGasServiceDataResponse
  {
      public int Id { get; set; }
      public int Version { get; set; }
  }
```


RepairsApi.Tests/E2ETests/GasServicingE2ETests.cs
```diff
@@ -86,6 +93,77 @@ public async Task GetDuplicates_WhenCalled_ReturnsGasDuplicates()
...
+            var id = _fixture.Create<int>();
+            var request = _fixture.Create<SaveGasServiceDataRequestObject>();
+
+            // Act
+            var response = await Post($"/api/v2/gasservicedata/{id}", request);
+
+            // Assert
+            response.Should().Be(HttpStatusCode.NotFound);
+        }
+
```
Asura5, line 166: Any case for a 400 test? Bad request for example has failed validation?

RepairsApi.Tests/V2/Controllers/GasServiceControllerTests.cs
```diff
@@ -138,5 +148,94 @@ public async Task FindAddressDuplicates_WhenCalled_ReturnsDuplicates()
...
+
+            var statusCode = GetStatusCode(response);
+
+            // Assert
+            statusCode.Should().Be((int) HttpStatusCode.Created);
+            responseContent.Should().BeEquivalentTo(useCaseResponse);
+        }
+
+        [Test]
+        public async Task Update_WhenCalled_ReturnsCreatedResponse()
```
Asura5, line 176: Should it return Created when you have not creating anything?

Asura5 (approved):

RepairsApi.Tests/E2ETests/GasServicingE2ETests.cs
LBHCallumM: Currently, the only validation is mapping the data to the request object. Unless you want me to validate each field?

PR #880, opened 2022-10-07
Separate gas servicing GetById and List routes

LBHCallumM:
## Describe this PR

I decided to refactor the Get route into two separate routes because it breaks the Single Responsibility Principle.


Asura5 (approved):
Approved, but would be keen to know what you think the responsibilities are here

LBHCallumM:
> Approved, but would be keen to know what you think the responsibilities are here

1. Get individual record with an id (if using id, there can only be one)
2. Get many records that match the search parameters

My thought process was, if the conditions change entirely based on the request id, then it should be its own method

PR #890, opened 2022-10-19
Callum/copy of daves circular reference branch

LBHCallumM:
## Describe this PR

The request object now takes a different Operative object:

```cs
 public class GasServiceOperativeRequestObject
    {
        public int Id { get; set; }
        public string Name { get; set; }
    }
```
I have included the Id, although it isn't currently used. 

The gateway methods will use the name to find the associated record. I'm not 100% happy with this approach, but a better approach would require more work. For example, integrating an operative lookup/dropdown into the table. This should work well if Don can type the operative name correctly.

The following scenarios could occur:
- Operative not found - returns bad request
- Multiple operatives found with the same name - returns bad request
-  Single matching operative found - success


Asura5 (approved):
Agree regarding the amount of work required for a substantial refactor - this is the best approach

PR #892, opened 2022-10-20
Update GasService endpoints to use operativeId instead of operativeName

LBHCallumM:
## Describe this PR

To get the operative lookup to work, it's much easier to just pass around the `operativeId`.


ange47rm (approved):
Approved, remove comment

PR #907, opened 2022-11-03
SOR Display Priority - Add column + populate

Asura5:
- Add a 'display_prioirty' column to the sor_codes table
- This is a non-nullable mandatory column but it has a default value of zero
- Zero value means "no specific display priority" 
- Greater than zero will mean "display at the top of the list", and vice versa


RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
```diff
@@ -183,6 +199,73 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+
+                    b.Property<int>("CountOfThirdAppts")
+                        .HasColumnType("integer")
+                        .HasColumnName("countofthirdappts");
+
+                    b.Property<int>("CountOfThirdNoAccesses")
+                        .HasColumnType("integer")
+                        .HasColumnName("countofthirdnoaccesses");
+
+                    b.ToTable((string)null);
```
ange47rm, line 248: No Table for this one?

RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
```diff
@@ -183,6 +199,73 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+
+                    b.ToView("gas_dashboard_analysis");
```
ange47rm, line 250: High chance I may have misunderstood this, but were we not going to have  a view containing GasBoilerAnalysis, GasExpiryAnalysis and GasDashboardAnalysis?

Asura5: Yep, that's correct. It's definitely worth you reading up on the EF Core naming conventions at https://github.com/efcore/EFCore.NamingConventions and our EF Core naming configuration - specifically Startup.cs:390 (use of UseSnakeCaseNamingConvention)

RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
Asura5: I'm not sure why that has been added to this auto-generated document - I think it's to ensure a table is not created (which is correct as it is a view - see line 250)

ange47rm (approved):

PR #916, opened 2022-11-11
MR-545 - Use of ReportType to fetch report specific data

ange47rm:
ReportType is now sent with requests to prompt the backend to fetch data specific for the report specified in the request.

Also added IsActive query param property to ListGasServiceDataRequest, should we need it in the future.

A check has been added in GasServiceController to check if the request includes the ReportType parameter. If yes, then we must make sure no other parameters are present (as ReportType is used as a unique parameter to retrieve report specific data and should not work in combination with other parameters). If other parameters, other than ReportType are present, a 400 Bad Request is returned to the client, otherwise the data for the reports is fetched.
Added tests:

- To check error 200 is returned if ReportType is sent by itself in the request.
- To check error 400 is returned if ReportType is sent in combination with other parameters.


RepairsApi/V2/Gateways/GasServiceGateway.cs
```diff
@@ -105,23 +127,36 @@ public async Task<List<GasServiceDataResponse>> ListGasServiceData(ListGasServic
                 query = query.Where(x => x.PostDate4 != null && ((DateTime) x.PostDate4).Date == ((DateTime) request.PostDate4).Date);
             }
 
-            if (request.Reason != null)
+            return query;
+        }
+
+        public IQueryable<GasServiceData> ListByReportType(ListGasServiceDataRequest request)
+        {
+            IQueryable<GasServiceData> query = _repairsContext.GasServiceData;
```
ange47rm, line 135: Not adding the 
```

                .Include(g => g.Operative1)
                .Include(g => g.Operative2)
                .Include(g => g.Operative3)
                .Include(g => g.Operative4);
```

would this cause an issue?

RepairsApi/V2/Boundary/Request/ListGasServiceDataRequest.cs
```diff
@@ -44,10 +53,14 @@ public class ListGasServiceDataRequest
         [FromQuery(Name = "reason")]
         public string Reason { get; set; }
 
+        [FromQuery(Name = "isactive")]
```
Asura5, line 56: What about if they specify a report type, and another parameter? Do you want to throw an error?

RepairsApi/V2/Gateways/GasServiceGateway.cs
Asura5: It depends - do your reports contain operative data?

ange47rm: No they don't, that was my reasoning that's why I decided to remove that

RepairsApi/V2/Boundary/Request/ListGasServiceDataRequest.cs
ange47rm: That's a good point

ange47rm: Addressed, PR description will be amended

RepairsApi/V2/Helpers/GasServiceDataGetRequestValidator.cs
```diff
@@ -0,0 +1,77 @@
+using RepairsApi.V2.Boundary.Request;
+using System;
+using System.Reflection;
+
+namespace RepairsApi.V2.Helpers
+{
+    public static class GasServiceDataGetRequestValidator
```
LBHCallumM, line 7: Can you add some tests for this?

RepairsApi/V2/Boundary/Request/ListGasServiceDataRequest.cs
LBHCallumM, line 56: `isActive`?

RepairsApi/V2/Boundary/Request/ListGasServiceDataRequest.cs
```diff
@@ -44,10 +53,14 @@ public class ListGasServiceDataRequest
...
+        public bool IsActive { get; set; } = true;
+
         [FromQuery(Name = "pageSize")]
         public int PageSize { get; set; } = 100;
 
-        [FromQuery(Name = "nullCompletionDate")]
-        public bool NullCompletionDate { get; set; } = false;
+
+        [FromQuery(Name = "reporttype")]
```
LBHCallumM, line 63: `reportType`?

RepairsApi/V2/Gateways/GasServiceGateway.cs
```diff
@@ -105,23 +127,36 @@ public async Task<List<GasServiceDataResponse>> ListGasServiceData(ListGasServic
...
+                query = query.Where(x => (x.Comments == "New Build") || (x.Comments == "New Build Non-Domestic Tightness Test") && (x.IsActive == false) && (x.CompletionDate == null) && (x.ExpiryDate < DateTime.Now));
+            }
 
-            _logger.LogInformation("About to get return {Count} results", results.Count());
-            return results.ToList();
+            if (request.ReportType == ReportType.VoidReport)
+            {
+                query = query.Where(x => (x.Comments == "Void Property") || (x.Comments == "Void Property-Gas Safety") && (x.IsActive == false));
+            }
+
```
LBHCallumM, line 158: Could you move the strings, such as "Gas Capped Off" into a class or enum? Or maybe an array.

Something like 

```cs
ValidVoidReportComments = new string[] {"Void Property", "Void Property-Gas Safety"}
```

LBHCallumM (commented):
Added a few comments

LBHCallumM (approved):

PR #922, opened 2022-11-17
Update list and addressDuplicate endpoints to enable pagination

LBHCallumM:
## Describe this PR

- List endpoint now includes 'pageNumber' parameter, which defaults to 1
- DuplicateAddress endpoint includes 'pagesize' and 'pageNumber' parameters
- The 'pageNumber' parameter starts at 1
- PageSize defaults to 25, as that is the number of results shown on the frontend

- Both endpoints now return `ListGasServiceDataRequest`

```javascript
{
  "pageNumber": 1,
  "totalResults": 24,
  "totalPages": 1,
  "data": [
    ... existingResponse
  ]
}
```

## Required Frontend Changes

- Accept different response objects for gasServiceData and addressDuplicates endpoints
- Handle serverside pagination from datagrid
- When fetching data for PDF's, pass max pageSize parameter (1000), and check for more results if pageNumber is less than totalPages


RepairsApi/V2/Boundary/Request/ListGasServiceDataRequest.cs
```diff
@@ -62,11 +62,13 @@ public class ListGasServiceDataRequest
...
+        public ReportType? ReportType { get; set; }
+
         [FromQuery(Name = "pagesize")]
         public int PageSize { get; set; } = 100;
 
-
-        [FromQuery(Name = "reporttype")]
-        public ReportType? ReportType { get; set; }
+        [FromQuery(Name = "page")]
+        public int Page { get; set; } = 0;
```
Asura5, line 72: Any reason why you've gone for a zero based page numbering system?

LBHCallumM: It's slightly less effort to implement (don't have to subtract 1). But I'm not too opinionated.

Asura5: The playbook is silent on this., so it's fine

RepairsApi/V2/Boundary/Response/ListGasServiceDataResponse.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+
+namespace RepairsApi.V2.Boundary.Response
+{
+    public class ListGasServiceDataResponse
+    {
+        public int Page { get; set; }
```
Asura5, line 7: You should stick to the Hackney standard when choosing the pagination query parameter names. See https://playbook.hackney.gov.uk/API-Playbook/naming_conventions#must-stick-to-conventional-query-parameters

LBHCallumM: For page and pageSize, I used the same as MMH

Asura5 (approved):

PR #928, opened 2022-11-25
Add RH security group roles to Gas groups

Asura5:
The google groups used to control access to our interim Gas Solution are:
- Gas-Servicing-Portal-Superuser
- Gas-Servicing-Portal-User
Through having one or both of these groups will allow access to the frontend, neither of these groups has any permissions mapped in RH, which is where we're getting our data
- That doesn't matter for the developers, because we're in loads of other groups as well, several of which give access to the /operatives endpoint (e.g. repairs-hub-operative-manager)
- Don and his team however are not in every group going, so they get a 401 when loading data tables (due to Authorize attributes being present on these endpoints)
- So, add mapping from the gas user groups to the internal_contractor role, which will open this endpoint, preventing 401 errors for Gas users


RepairsApi/V2/Infrastructure/Migrations/20221125145635_AddGasRolesToSecurityGroups.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace RepairsApi.V2.Infrastructure.Migrations
+{
+    public partial class AddGasRolesToSecurityGroups : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.InsertData(
+                table: "security_groups",
+                columns: new[] { "group_name", "user_type", "id" },
+                values: new object[,]
```
ange47rm, line 14: Out of curiosity, is the comma in the [ ] as a delimiter?

Asura5: In c# that's the declaration for a two dimensional array (of anything, in this case Objects):
[] = 1d array
[,] = 2d array
If you look at the initialization, you can see I'm inserting two rows of 3 columns, so an array of <an array of strings>

ange47rm (approved):

PR #931, opened 2022-11-29
Improve Address Search

LBHCallumM:
## Describe this PR

Add normalisation to address search, including:
- Setting to lowercase
- Removing excess space

Add `DateMatches` helper method for date comparison.


Asura5 (approved):
LGTM but you might get an EF error due to the function call being untranslatable into SQL (when the view is materialised)

PR #935, opened 2022-12-01
MR-604 - Add Boiler Types

ange47rm:
Follow up to PR https://github.com/LBHackney-IT/gas-servicing-frontend/pull/93

This PR adds Boiler Type property to Gas Service Data

(TBC)


RepairsApi/V2/Boundary/Response/GasServiceDataResponse.cs
```diff
@@ -55,6 +55,7 @@ public class GasServiceDataResponse
 
         public DateTime? ForcedEntryDate { get; set; }
 
+        public virtual GasBoilerType Type { get; set; }
```
LBHCallumM, line 58: Have you tested that this data is returned via the getById/List endpoints? For the operatives, I had to include:

```cs
.Include(g => g.Operative1)
```

ange47rm: I can see the Type is returned in the same way Make and Model are, is this what you mean?
image

PR #936, opened 2022-12-05
Populate SorCode Description change for Repairs Online

LBHCallumM:
## Link to JIRA ticket

[Ticket](https://hackney.atlassian.net/browse/MR-605?atlOrigin=eyJpIjoiNzAzMjQ4MWNhNjgwNDJlMzgyNWVlMGZiNGQzYzU3ODUiLCJwIjoiaiJ9)

## Describe this PR

When the value of `CustomName` is null, populate the value from the database.

- If `CustomName` is null, but sorCode is not found, throws an exception since the description cannot be populated.


RepairsApi/V2/Gateways/RepairsGateway.cs
```diff
@@ -41,13 +46,44 @@ public async Task<int> CreateWorkOrder(WorkOrder workOrder)
...
             await _repairsContext.SaveChangesAsync();
 
             return entry.Entity.Id;
         }
 
+        private async Task PopulateSorDescriptions(WorkOrder workOrder)
+        {
+            foreach (var workElement in workOrder.WorkElements)
+            {
+                var rsi = workElement.RateScheduleItem.First();
```
Asura5, line 66: Single()? Do we care if there are multiple?

Asura5 (approved):

PR #956, opened 2022-12-22
Increase raiselimit and spendlimit to 1000 for m2m users

LBHCallumM:
## Describe this PR

When raising a workOrder from Repairs Online Staging with an M2M token, the workOrder status is set to 'Pending Approval'. This is because the raiseLimit was 0.

I have increased the vary limit to 1000. I matched the vary limit assigned to the operative security group.

image


Asura5 (commented):
It might be worth checking the maximum value of a repair from repair online to ensure 1000 is high enough. I can do this after lunch if you need but until then it's sorconfig.json in the HROL BE API

LBHCallumM:
> It might be worth checking the maximum value of a repair from repair online to ensure 1000 is high enough. I can do this after lunch if you need but until then it's sorconfig.json in the HROL BE API

The highest cost I can find is £101, so I think 1k is ok

ange47rm (approved):

PR #964, opened 2023-01-13
Update GetTrades and GetContractors endpoints

LBHCallumM:
Update GetTrades and GetContractors endpoints to accept optional parameters. However, if `GetAllContractors` or `GetAllTrades` is false, the other parameters are still required.


RepairsApi/V2/Controllers/ContractorController.cs
```diff
@@ -22,21 +24,29 @@ public ContractorController(IScheduleOfRatesGateway sorGateway)
...
-            var contractors = await _sorGateway.GetContractors(propertyReference, tradeCode);
+            if (!request.GetAllContractors && string.IsNullOrWhiteSpace(request.TradeCode))
+            {
+                return BadRequest($"If {nameof(request.GetAllContractors)} is False, you must include a value for {nameof(request.TradeCode)}");
+            }
 
-            return Ok(contractors);
+            if (!request.GetAllContractors && string.IsNullOrWhiteSpace(request.PropertyReference))
+            {
+                return BadRequest($"If {nameof(request.GetAllContractors)} is False, you must include a value for {nameof(request.PropertyReference)}");
```
Asura5, line 41: What if they include both?

RepairsApi/V2/Controllers/ScheduleOfRatesController.cs
```diff
@@ -103,12 +103,21 @@ public async Task<IActionResult> GetSorCode([Required] string sorCode, [FromQuer
...
+        [Route("trades")]
+        [ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(typeof(IEnumerable<SorTradeResponse>), StatusCodes.Status200OK)]
+        [ProducesDefaultResponseType]
+        public async Task<IActionResult> ListTrades(ListTradesRequestObject request)
         {
-            return Ok(await _listSorTrades.Execute(propRef));
+            if (!request.GetAllTrades && string.IsNullOrWhiteSpace(request.PropertyReference))
+            {
+                return BadRequest($"If {nameof(request.GetAllTrades)} is False, you must include a value for {nameof(request.PropertyReference)}");
```
Asura5, line 115: What if they include both?

RepairsApi/V2/Controllers/ContractorController.cs
LBHCallumM: Both as in propRef and getAllContractors? GetAllContractors returns all contractors, so it doesn't need the propRef. Its the same for trades

Asura5: So why is there not validation disallowing a request where both of them are supplied

PR #973, opened 2023-01-23
Add trunc warning to order comments

Asura5:
- When creating a RHOL work order
- Cyndi is worried that planners will miss vital info
- Add warning indicating that the field is truncated and they have access to Repairs hub
- See Sean's context here: https://hackney.atlassian.net/browse/MR-655


RepairsApi/V2/Services/DRS/DrsMapping.cs
```diff
@@ -63,7 +63,12 @@ private async Task<order> CreateOrder(WorkOrder workOrder, bool ignoreContractVa
 
             var uniqueCodes = locationAlerts?.Union(personAlerts);
             var orderComments =
-                @$"{uniqueCodes.ToCodeString()} {workOrder.DescriptionOfWork}".Truncate(230);
+                @$"{uniqueCodes.ToCodeString()} {workOrder.DescriptionOfWork}";
+
+            if (orderComments.Length > 216)
```
ange47rm, line 68: This way we're basically making sure that we truncate after the 216th char, and add " <SEE REP HUB>", which makes it 229 chars in total. 
Didn't they say they tested it, and found out 125 is the max number of char's allowed, or are we sure that as long as we send less than 255 we're good (which makes perfect sense)?

Asura5: It makes it 230 not 229. The documentation says 255 but our empirical testing suggests 230 which is why we've limited it to that

Asura5: It's not 125 total it's apparently 255 total (in reality 230 total), so (desc + alerts + HROL issue) < 230. Sean's comment explains it well except internally it appears 230 is the number rather than 255

ange47rm: Ok sure, I just wanted to double check. The way I had it in my head was as if someone had tested all the possible amount of chars and ended up finding out 125 was the max number.

ange47rm (approved):
Approved, however I would suggest reviewing the Sonarcloud possible DoS related issue/vulnerability before merging. 

Line 107 in DrsMapping.cs

`Not specifying a timeout for regular expressions can lead to a Denial-of-Service attack. Pass a timeout when using System.Text.RegularExpressions to process untrusted input because a malicious user might craft a value for which the evaluation lasts excessively long.`

PR #983, opened 2023-02-14
Fix syntax when calling stored procedure

LBHCallumM:
Change parameters used when calling stored procedure. The following error appeared in Sentry.

image


RepairsApi/V2/Gateways/BackOfficeGateway.cs
```diff
@@ -104,7 +104,7 @@ public async Task<int> AddContractToProperty(BackOfficeSorContractsRequestObject
 
             try
             {
-                await _repairsContext.Database.ExecuteSqlRawAsync("CALL sp_addpropertycontract(@dest, @ctr);", dest, ctr);
+                await _repairsContext.Database.ExecuteSqlRawAsync("CALL sp_addpropertycontract(@destinationpropref, @sourcepropref);", dest, ctr);
```
ange47rm, line 107: Where do these "correct" parameters come from?

LBHCallumM: They are defined when creating `NpgsqlParameter`'s

ange47rm (approved):

PR #987, opened 2023-03-28
Use "as at date"

Asura5:
- Add a contract validity date to the SOR request pipeline
- Defaults to the old behaviour, which was current Datetime
- USeful for testing and simulations


RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
```diff
@@ -105,7 +105,7 @@ public async Task<IEnumerable<ScheduleOfRatesModel>> GetSorCodes(ListScheduleOfR
 
         public async Task<IEnumerable<ScheduleOfRatesModel>> CheckSorCodes(CheckSorCodesRequest request)
         {
-            var query = BaseGetAllSorCodes(request.PropertyReference, request.ContractorReference);
+            var query = BaseGetAllSorCodes(request.PropertyReference, DateTime.UtcNow, request.ContractorReference);
```
LBHCallumM, line 108: You could pass null instead of the date, since it's not going to be used.

LBHCallumM (approved):

PR #989, opened 2023-03-31
Update staging

Asura5:


ange47rm (approved):

RepairsApi/V2/Gateways/ScheduleOfRatesGateway.cs
```diff
@@ -129,14 +129,16 @@ public async Task<IEnumerable<ScheduleOfRatesModel>> CheckSorCodes(CheckSorCodes
             }).ToListAsync();
         }
 
-        private IQueryable<ScheduleOfRatesDTO> BaseGetAllSorCodes(string propertyReference, string contractorReference = null)
+        private IQueryable<ScheduleOfRatesDTO> BaseGetAllSorCodes(string propertyReference, DateTime? contractEffectiveDate, string contractorReference = null)
         {
+            var ced = contractEffectiveDate.HasValue ? contractEffectiveDate.Value : DateTime.UtcNow;
```
ange47rm, line 134: Could we not have set contractEffectiveDate to be DateTime.UtcNow as default in the method signature instead of having a ternary operator and a new 'ced' variable?

PR #999, opened 2023-04-26
Upgraded version of Hackney.Asset.Shared to 0.30

ange47rm:
## Link to JIRA ticket

MR-698 - https://hackney.atlassian.net/jira/software/c/projects/MR/boards/88?modal=detail&selectedIssue=MR-698

While working on the New Asset Form (MMH) we've noticed that, when saving a new asset, a lot of other properties are added to it.
An asset can have so many properties, however the New Asset Form only captures some of these. The issue is with the properties we're not capturing, especially the ones of type int and boolean, as a default value is assigned, which can be misleading for example:
`numberOfShowers` is saved with a `0` value or a boolean property such as `hasRampAccess` is saved as `false` - these are fields the values of which we're NOT capturing in the form, but if someone was to look at the asset in the database, they would find misleading information.
For the above examples, as we're not capturing those properties (or they're optional in the 'New asset' form), we'd want their values to be null, rather than use default values (0s and false's).

- Reviewed where Hackney.Asset.Shared is used in this repo
- Upgraded version of Hackney.Asset.Shared to 0.30

### *Changes introduced by in Hackney.Asset.Shared*
Properties within the AssetDb class (within Hackney.Asset.Shared) of type 'int' and 'boolean' have been set from non-nullable to nullable with the latest update of the library, to version 0.30.


ange47rm:
Closing PR. Solution updated by PR #1012.

PR #1000, opened 2023-04-28
Update StartTime in DRS

LBHCallumM:
## Summary of Changes
- when calculating a booking startTime, the value from the workOrder is always default (if not null)
- Added test to DrsMappingTest that asserts the workOrder startTime is used. (for other tests, the startTime is set as null)
- tidied up tests (+ removed an unused method)
- tidied up interface IDrsService


RepairsApi.Tests/V2/Services/DRS/DrsMappingTests.cs
```diff
@@ -107,77 +103,118 @@ public void Setup()
...
+            workOrder.StartTime = startTime;
+
+            // Act
+            var drsOrderCopy = drsOrder.DeepClone();
+            var request = await _classUnderTest.BuildStartOrderUpdate(SessionId, workOrder, drsOrderCopy);
+
+            // Assert
+            foreach (var updateBooking in request)
+            {
+                ValidateTargetDate(startTime, updateBooking.updateBooking1.startDateAndTime);
```
Asura5, line 192: Is "ValidateTargetTime" still the right thing to call this method? Originally it was used to validate an attribute called  "target time" (i..e estimated completion time)

Asura5 (approved):

PR #1015, opened 2023-06-26
Update DRS ClosedDate logic

LBHCallumM:
## Describe this PR

I was assigned a Halo ticket from Cyndi, saying she had noticed that since the start time feature had been released, closedTimes in DRS were incorrect.

I was able to replicate this issue. However, more interestingly, I have found workOrders from before that release with the same issue. Although the closedTimes are incorrect on DRS, they are correctly saved to the `WorkOrder` and `WorkOrderCompletion` tables.

Upon further inspection of `DrsNotificationHandler`, only the `workOrder`, and therefore `workOrder.ClosedDate` is used when updating DRS. `CompleteWorkOrderUseCase` only updates `workOrder.ClosedDate` after DRS is updated. 

To ensure the correct ClosedDate is sent to DRS, I have moved the logic that updates `workOrder.ClosedDate` before DRS is updated.


RepairsApi/V2/UseCase/CompleteWorkOrderUseCase.cs
```diff
@@ -95,6 +95,9 @@ private async Task UpdateWorkOrderStatus(WorkOrder workOrder, WorkOrderComplete
 
             workOrderComplete.ClosedTime ??= update.EventTime;
 
+            _logger.LogInformation("Setting workOrder.ClosedDate to {ClosedTime} ?? {EventTime} for {WorkOrderId}", workOrderComplete.ClosedTime, update.EventTime, workOrder.Id);
```
Asura5, line 98: Not sure you need the null coalesce given you have line 96

Asura5 (approved):

RepairsApi/V2/UseCase/CompleteWorkOrderUseCase.cs
LBHCallumM: I added this so I could determine which value is used on line 99.

PR #1020, opened 2023-07-12
Add getPropertyByAssetId endpoint

LBHCallumM:
## Ticket

[https://hackney.atlassian.net/browse/MR-716?atlOrigin=eyJpIjoiYTBmYWIyMzk3YzMwNDQ3ZWJmN2MzOGZhZGI5YWY4ZjYiLCJwIjoiaiJ9](https://hackney.atlassian.net/browse/MR-716?atlOrigin=eyJpIjoiYTBmYWIyMzk3YzMwNDQ3ZWJmN2MzOGZhZGI5YWY4ZjYiLCJwIjoiaiJ9)

## Summary of Changes

For the boilerHouse integration, the getProperty endpoint now returns a boilerHouseId. However, this is a GUID, not a propertyReference (which is used in repairs). 

For the boilerHouse feature to work, I need to grab property information for that property using the GUID. 

In this PR, I have added a new endpoint to RepairsAPI that can fetch property information using the GUID provided.

### Key changes

- Many previous methods that were called `getProperty*` have been renamed to `getPropertyByPropertyReference`
- Refactoring of some logging


RepairsApi/V2/Controllers/PropertiesController.cs
```diff
@@ -22,22 +22,25 @@ namespace RepairsApi.V2.Controllers
...
+            IGetPropertyByAssetIdUseCase getPropertyByAssetIdUseCase,
             IHousingSearchUseCase housingSearchUseCase,
             ILegalDisrepairUseCase legalDisrepairUseCase,
             IPersonAlertsUseCase getPersonAlertsUseCase,
             ILocationAlertsUseCase getLocationAlertsUseCase,
             ILogger<PropertiesController> logger)
         {
-            _getPropertyUseCase = getPropertyUseCase;
+            _getPropertyByPropertyReferenceUseCase = getPropertyByPropertyReferenceUseCase;
+            _getPropertyByAssetIdUseCase = getPropertyByAssetIdUseCase;
```
Asura5, line 43: I think this is confusing and you should change it. AssetId is the name of the index on Dynamo which returns properties by the 8/9 digit property reference:

image

However here you are referring to the Guid, which is Id in Dynamo. I think we should avoid he use of AssetId altogether and be explicit e.g. getPropertyByGuidPK or getPropertyByGuid

LBHCallumM: Good spot, I agree

LBHCallumM: Updated

Asura5 (approved):

PR #1028, opened 2023-07-25
Add saltash sec group

Asura5:
We have a request for a search option for Saltash ([https://hackney.haloitsm.com/ticket?id=101094](https://hackney.haloitsm.com/ticket?id=101094))

For a contractor to appear in the search menu on the left hand side of the dashboard, the following needs to happen:

1. A security group needs to be created [in this case repairs-hub-contractors-saltash]
2. A DB migration against the security groups table needs to happen, containing the matching group_name (this PR)
3. The new user is then added to this Google group


RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
```diff
@@ -1626,6 +1626,10 @@ protected override void BuildModel(ModelBuilder modelBuilder)
                         .HasColumnType("integer")
                         .HasColumnName("site_id");
 
+                    b.Property<DateTime?>("StartTime")
+                        .HasColumnType("timestamp without time zone")
+                        .HasColumnName("start_time");
+
```
LBHCallumM, line 1632: Was anything else modified? Im not sure why this was added to the snapshot.

LBHCallumM (approved):
Added a comment

RepairsApi/V2/Infrastructure/Migrations/RepairsContextModelSnapshot.cs
Asura5: nothing else was modified, but that column does exist - it may have been an inaccurate former migration

PR #1029, opened 2023-07-28
Trim operative resource code

Asura5:
- The operative IDs used to sync work orders from DRS are based on the employee ID supplied in DRS
- The users of DRS frequently copy and paste from sheets when creating these resources and shock horror, there's no validation on DRS inputs. This leads to SOAP responses like:
image

So, trim the employee ID of all whitespace

Root HALO: https://hackney.haloitsm.com/ticket?id=100297


LBHCallumM (approved):

RepairsApi.Tests/V2/Services/DRS/DrsServiceTests.cs
```diff
@@ -16,6 +16,7 @@
 using RepairsApi.V2.Exceptions;
 using RepairsApi.V2.Gateways;
 using RepairsApi.V2.Services.DRS;
+using Amazon.Auth.AccessControlPolicy;
```
Duslerke, line 19: Is this import being used by the new code added?
I can't seem to find anything obvious using it.

PR #1034, opened 2023-08-23
MR-516 - Edit stored procedure to add contracts to new properties

ange47rm:
Edit stored procedure to add contracts to new properties

When the repairs-listener picks up a new message related to the creation of a New Asset, this stored procedure, when/if run, adds all or any missing contracts to the new property (in those cases when the propRef may already be in the Repairs DB).


RepairsApi/V2/Infrastructure/Migrations/20230823143558_EditStoredProcAddContractsToNewAssets.cs
```diff
@@ -0,0 +1,74 @@
...
+                    ON CONFLICT (prop_ref, contract_reference) DO NOTHING;
+	
+                end; $$
+            ");
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.Sql(@"
+                CREATE OR REPLACE PROCEDURE assign_dlo_property_contracts_to_newbuild(
```
ange47rm, line 38: Set back to original/pre-PR version

Asura5 (approved):

PR #1035, opened 2023-08-30
One Job at a Time - Database

Asura5:
- One job at a time is per operative
- Default to false (for now) to maintain existing experience


RepairsApi/V2/Infrastructure/Migrations/20230830140647_AddOneJobAtATimeDbColumns.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            migrationBuilder.AddColumn<bool>(
+                name: "is_onejobatatime",
+                table: "operatives",
+                type: "boolean",
+                defaultValue: false,
+                nullable: true
+                );
+
+            migrationBuilder.Sql("UPDATE public.operatives SET is_onejobatatime = true WHERE email = 'repairs-hub-test2@hackney.gov.uk'");
```
Duslerke, line 20: I've got no context for what this PR is about, so I doubt I can do 
That being said: this email seems like a test email. Is this intended to be hardcoded or be this particular email in general?

Asura5: Yep, that's right. We're going to enable this feature just for one user to start with

Asura5: (This user exists only in staging, so the statement will have no impact on other environments. This is also intentional)

Duslerke (approved):
Haven't read through the `Designer` file. as my understanding is it's something that system generates & consumes. _(The actual reviewable changes are the other file)_

The other migration file seems good for both Up and Down migrations.

PR #1036, opened 2023-08-30
Surface one job at a time up to controller

Asura5:
Now it's in the database, it's time to party


RepairsApi/V2/Boundary/Response/OperativeResponse.cs
```diff
@@ -8,5 +8,6 @@ public class OperativeResponse
         public string PayrollNumber { get; set; }
         public string Name { get; set; }
         public List<string> Trades { get; set; }
+        public bool IsOnejobatatime { get; set; }
```
ange47rm, line 11: Can you please change it to IsOneJobAtATime 🙏 , unless you really dislike that

ange47rm (approved):
image

ange47rm:
image

RepairsApi/V2/Boundary/Response/OperativeResponse.cs
Asura5: Unfortunately not, because that would need a database column called is_one_job_at_a_time which looks equally filthy

Asura5: It's also for consistency - we have IsArchived (mapped to is_archived) on the the same table

ange47rm: Ah ok no probs

PR #1040, opened 2023-09-18
Property endpoint contact details optional

LBHCallumM:
## Summary of Changes

Contact details are returned from the property endpoint on repairs. However, the frontend only uses contact details provided by the property endpoint when raising a repair.

Due to the implementation, the contact details gateway is very slow. A request needs to be made for each household member. This means if there are 4 household members, four additional requests are made when the data probably isn't required.

To resolve this, I have added a query parameter to the endpoint `withContactDetails` (which defaults to false). Before this PR is merged, I will update the frontend, so that the flag is passed as `true` when raising a repair.


RepairsApi/V2/UseCase/GetPropertyByPropertyReferenceUseCase.cs
```diff
@@ -41,17 +43,26 @@ public async Task<PropertyBasicDetails> ExecuteAsync(string propertyReference)
...
-            var contactDetails = await _contactDetailsGateway.GetContactDetailsAsync(tenureInformation?.HouseholdMembers);
-
             _logger.LogInformation("All information for asset with {PropertyReference} has been successfully gathered. {AssetResponse}. Returning PropertyBasicDetails object. ", propertyReference, JsonConvert.SerializeObject(assetResponse));
 
             return new PropertyBasicDetails
             {
                 PropertyModel = assetResponse.ToDomain(),
                 Tenure = tenureInformation,
-                ContactDetails = contactDetails
+                ContactDetails = await GetContactDetails(withContactDetails, propertyReference, tenureInformation?.HouseholdMembers)
```
ange47rm, line 52: Just being picky, not saying we need to make changes. This calls GetContactDetails even if withContactDetails is `false`.

Could line 52 maybe be turned into the following?
`ContactDetails = withContactDetails ? await GetContactDetails(propertyReference, tenureInformation?.HouseholdMembers) : null`?

This way you could then remove line 58 completely

ange47rm:
When we merge this PR contact details won't be passed unless explicitly requested with the boolean value. That's ok.

I see you've added PR https://github.com/LBHackney-IT/repairs-hub-frontend/pull/772 - Could there be other systems/features/pages in RH that need changing as they rely on those contact details to be returned?

ange47rm (approved):

RepairsApi/V2/UseCase/GetPropertyByPropertyReferenceUseCase.cs
LBHCallumM: Do you not think your suggestion is harder to read?

PR #1043, opened 2023-09-20
Add contact details endpoint

LBHCallumM:
## Summary of Changes

I was unable to get the previous implementation due to E2E tests. That approach was to implement the parameter `withContactDetails` into the existing property endpoint. However, due to how Cypress intercepts work, this change became too difficult to resolve.

[https://github.com/LBHackney-IT/repairs-hub-frontend/pull/772](https://github.com/LBHackney-IT/repairs-hub-frontend/pull/772)

Instead, I will take the following approach:
1. Add a new contact details endpoint
2. Update the frontend to use the contact details endpoint
3. Remove contact details from the property endpoint

## Amendment

Add tenureId to property response. Needed to fetch contact details.


RepairsApi.Tests/E2ETests/GetContactDetailsE2ETests.cs
```diff
@@ -0,0 +1,88 @@
...
+        private readonly Fixture _fixture = new Fixture();
+
+        [SetUp]
+        public void SetUp()
+        {
+            MockApiGateway.ForcedCode = null;
+        }
+
+        [Test]
+        public void WhenTenureNotFound_ReturnsNotFoundResponse()
```
Asura5, line 34: copy and paste error?

RepairsApi.Tests/E2ETests/GetContactDetailsE2ETests.cs
```diff
@@ -0,0 +1,88 @@
...
+            response.Status.Should().Be(HttpStatusCode.NotFound);
+        }
+
+        [Test]
+        public void WhenContactDetailsFound_ReturnsContactDetails()
+        {
+            // Arrange
+            var client = new ApiGateway(new HttpClientFactoryWrapper(Client));
+
+            var tenureId = _fixture.Create<string>();
```
Asura5, line 57: Why does this need to be a random string - it's not used anywhere except the method call

LBHCallumM: It doesnt need to be random. I prefer this approach because it avoids using magic strings. I can use a hardcoded string if you think it makes it more readable

Asura5 (approved):

PR #1045, opened 2023-09-26
Remove contact details from properties endpoint

LBHCallumM:


ange47rm (approved):

ange47rm (approved):
The changes are clear, when we get a property now we don't automatically get the related contact details. 

Have you taken into considerations how other solutions that use RepairsAPI's GetProperty endpoint will react to the lack of contact details? Mostly/mainly RH? I think adding a PR description could have helped here

PR #1058, opened 2023-10-30
Remove Gas Service schema from RepairsHub DB

Asura5:
- Remove views
- Remove indexes
- Remove tables


RepairsApi.Tests/RepairsApi.Tests.csproj
```diff
@@ -68,6 +68,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Folder Include="E2ETests\Helpers\" />
```
adamtry, line 71: Should this be included?

adamtry (approved):
Looks good to me - did you mean to include the helpers folder in the tests csproj?
Left a comment for that

RepairsApi.Tests/RepairsApi.Tests.csproj
Asura5: there was only one file in that folder, which was a gas helper class

PR #1070, opened 2023-12-21
Get person alerts by tenureid instead of tenancy ref

adamtry:
[MR-872](https://hackney.atlassian.net/jira/software/c/projects/MR/boards/88?selectedIssue=MR-872)

Updating the person-alerts endpoint to take the tenure ID instead of a tenancy agreement reference so that it can fetch the tenure (tenure API) and then all cautionary alerts for each person (by person ID) in the tenure, which will give more reliable + up to date data.

[MR-872]: https://hackney.atlassian.net/browse/MR-872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RepairsApi/V2/UseCase/PersonAlertsUseCase.cs
```diff
@@ -3,21 +3,44 @@
...
         {
             _alertsGateway = alertsGateway;
+            _tenureGateway = tenureGateway;
         }
 
-        public async Task<BasicAlertList> ExecuteAsync(string tenancyAgreementReference)
+        public async Task<BasicAlertList> ExecuteAsync(string tenureId)
         {
-            return await _alertsGateway.GetPersonAlertsAsync(tenancyAgreementReference);
+            var tenure = _tenureGateway.GetById(tenureId).Result;
```
LBHCallumM, line 23: Is there any reason you've used `.Result` instead of `await`. I could be wrong, but I'm pretty sure this approach will block the thread for any other requests.

adamtry: I had a look as well and seems it does.
I switched it back to await in 11d79254a36b46998dc2ddc9435cd92f35fb86c0

RepairsApi/V2/UseCase/PersonAlertsUseCase.cs
```diff
@@ -3,21 +3,44 @@
...
+            {
+                throw new ResourceNotFoundException(tenureId);
+            }
+
+            var alerts = new List<Alert>();
+            foreach (var member in tenure.HouseholdMembers)
+            {
+                var memberAlerts = await _alertsGateway.GetPersonAlertsAsync(member.Id.ToString());
+                alerts.AddRange(memberAlerts.Alerts);
+            }
```
LBHCallumM, line 35: Since there can be multiple contacts, you can easily speed this up by waiting for each request simultaneously by using `Task.WhenAll`. 

For reference, its been used in `ContactDetailsGateway`.

LBHCallumM: Also, `GetPersonAlertsAsync` is returning the `BasicAlertsList` when only the alerts are being used. Could you return the list directly instead?

Secondly, you can replace line 37 with a call to the factory method in `GetPersonAlertsAsync`: ` return response.Content.ToDomain(tenancyReference);`

RepairsApi/V2/Services/DRS/DrsMapping.cs
```diff
@@ -125,15 +125,20 @@ private async Task<IEnumerable<Alert>> GetPersonAlerts(TenureInformation tenureI
                 return new List<Alert>();
             }
 
-            var result = await _alertsGateway.GetPersonAlertsAsync(tenureInformation.TenancyAgreementReference);
+            var alerts = new List<Alert>();
+            foreach (var member in tenureInformation.HouseholdMembers)
+            {
+                var memberAlerts = await _alertsGateway.GetPersonAlertsAsync(member.Id.ToString());
+                alerts.AddRange(memberAlerts.Alerts);
+            }
```
LBHCallumM, line 133: Same comment about `Task.WhenAll`

RepairsApi/V2/Services/DRS/DrsMapping.cs
```diff
@@ -125,15 +125,20 @@ private async Task<IEnumerable<Alert>> GetPersonAlerts(TenureInformation tenureI
...
-            if (result == null)
+            if (alerts.Count == 0)
             {
-                _logger.LogWarning("GetPersonAlertsAsync returned null for {TenancyAgreementReference}", tenureInformation?.TenancyAgreementReference);
+                _logger.LogWarning("GetPersonAlertsAsync returned no alerts for {TenancyAgreementReference}", tenureInformation?.TenancyAgreementReference);
                 return new List<Alert>();
             }
 
-            return result.Alerts;
+            return alerts;
```
LBHCallumM, line 141: I know that you didn't write this, but this logic doesnt make sense. We know there are alerts, so why create a new array instead of returning existing list (which is empty). 

Maybe just log the number of alerts and remove the condition?

LBHCallumM (approved):
LGTM - Just added a few suggestions

PR #1075, opened 2024-01-10
POST endpoint for collecting damp and mould reports

adamtry:
[MR-869](https://hackney.atlassian.net/jira/software/c/projects/MR/boards/88?selectedIssue=MR-869) As a BSO, I need operatives to report damp and mould when completing a Repair

This PR is for the endpoint that allows the creation of new damp and mould reports.
Depends on the creation of a new table in the database which is generated with this SQL:

```sql
CREATE TABLE IF NOT EXISTS "damp_and_mould_reports" (
  Id SERIAL PRIMARY KEY NOT NULL,
  property_reference TEXT NOT NULL,
  damp_and_mould_present BOOLEAN NOT NULL,
  previously_reported BOOLEAN NOT NULL,
  previous_report_resolved BOOLEAN NOT NULL,
  comments TEXT,
  reported_at TIMESTAMP NOT NULL DEFAULT NOW()
);
```

`previously_reported` and `previous_report_resolved` default to false in the API if not provided

New POST endpoint created: `/api/v2/damp-and-mould/reports/:propertyRef`, takes request object from client and returns NoContent.

The plan is to follow this with one or more GET endpoints to fetch the created entities.


mr-kg (approved):

RepairsApi/V2/Infrastructure/DampAndMouldReport.cs
```diff
@@ -0,0 +1,22 @@
...
+        public string PropertyReference { get; set; }
+        [Required]
+        public bool DampAndMouldPresent { get; set; }
+        [Required]
+        public bool PreviouslyReported { get; set; }
+        [Required]
+        public bool PreviousReportResolved { get; set; }
+        public string Comments { get; set; }
+        [Required]
+        public DateTime? ReportedAt { get; set; }
```
LBHCallumM, line 20: Does ReportedAt need to be nullable? When displaying the reports, were going to want to show them newest first. If the value is null, there is no way to order it correctly.

RepairsApi/V2/Controllers/DampAndMouldController.cs
```diff
@@ -0,0 +1,58 @@
...
+        public async Task<IActionResult> ReportDampAndMould([FromRoute] string propertyReference, [FromBody] DampAndMouldReportRequest request)
+        {
+            if (!ModelState.IsValid || request == null)
+            {
+                return BadRequest(ModelState);
+            }
+
+            propertyReference = propertyReference.Trim();
+            var isDigitRegex = new Regex(@"^\d+$", RegexOptions.None, TimeSpan.FromMilliseconds(1000));
+            if (!isDigitRegex.IsMatch(propertyReference))
```
LBHCallumM, line 49: There is quite a lot going on here. Could we extract this into a method, called something like `validatePropertyReference`?

RepairsApi/V2/Controllers/DampAndMouldController.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// <param name="request"></param>
+        /// <param name="propertyReference"></param>
+        [HttpPost]
+        [Route("reports/{propertyReference}")]
+        [ProducesResponseType(StatusCodes.Status204NoContent)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(typeof(string), StatusCodes.Status403Forbidden)]
+        [ProducesDefaultResponseType]
+        [Authorize(Roles = UserGroups.Contractor + "," + UserGroups.ContractManager)]
+        public async Task<IActionResult> ReportDampAndMould([FromRoute] string propertyReference, [FromBody] DampAndMouldReportRequest request)
```
LBHCallumM, line 40: This is nit-picking, and probably not worth changing at this point. Would it be easier to pass the propertyReference in the body? I cant see a reason to include it in the url, as its not referencing a related entity. Im not 100% sure on the rules for REST apis.

RepairsApi.Tests/V2/UseCase/ReportDampAndMouldUseCaseTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            // Assert
+            _dampAndMouldGatewayMock.Verify(gw => gw.AddDampAndMouldReport(It.IsAny<DampAndMouldReport>()), Times.Once);
+            _dampAndMouldGatewayMock.Verify(
+                gw => gw.AddDampAndMouldReport(It.Is<DampAndMouldReport>(r =>
+                    r.DampAndMouldPresent == request.DampAndMouldPresent &&
+                    r.PreviouslyReported == request.PreviouslyReported &&
+                    r.PreviousReportResolved == request.PreviousReportResolved &&
+                    r.Comments == request.Comments &&
+                    r.ReportedAt != null
+                )), Times.Once);
```
LBHCallumM, line 54: This is testing both the gateway being called, and the factory method. I would suggest moving the second test to the factory class tests file.

Could you instead use factory method within the test? For example:
```cs
_dampAndMouldGatewayMock.Verify(gw => gw.AddDampAndMouldReport(report.ToDb(propRef)), Times.Once);
```

RepairsApi/V2/Gateways/DampAndMouldGatway.cs
```diff
@@ -0,0 +1,27 @@
+using RepairsApi.V2.Infrastructure;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using RepairsApi.V2.Gateways.Interfaces;
+
+namespace RepairsApi.V2.Gateways
+{
+    public class DampAndMouldGateway : IDampAndMouldGateway
```
LBHCallumM, line 8: Just noticed. the filename is missing an e :)

RepairsApi/V2/Infrastructure/DampAndMouldReport.cs
adamtry: True. The database defaults to NOW() if it is null, but best to have the model essentially reflect this

Fixed in 730afef9095cb00f9a6c399c228fe4d56062fe2b

RepairsApi/V2/Gateways/DampAndMouldGatway.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public DampAndMouldGateway(RepairsContext repairsContext, ILogger<DampAndMouldGateway> logger)
+        {
+            _repairsContext = repairsContext;
+            _logger = logger;
+        }
+
+        public async Task<int> AddDampAndMouldReport(DampAndMouldReport report)
+        {
+            _logger.LogInformation($"Adding damp and mould report");
```
LBHCallumM, line 21: Could we include the propRef in this log to make it more identifiable?

RepairsApi.Tests/V2/Controllers/DampAndMouldControllerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            // assert
+            result.Should().BeOfType<NoContentResult>();
+        }
+
+        [Test]
+        public async Task ReportDampAndMould_WhenRequestInvalid_ReturnsBadRequest()
+        {
+            // arrange
+            _reportDampAndMouldUseCase.Setup(m =>
+                m.ExecuteAsync(It.IsAny<DampAndMouldReportRequest>(), It.IsAny<string>()));
```
LBHCallumM, line 51: I dont think you need to setup the mock for this test

RepairsApi.Tests/V2/Controllers/DampAndMouldControllerTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using RepairsApi.V2.Boundary.Request;
+
+namespace RepairsApi.Tests.V2.Controllers
+{
+    public class DampAndMouldControllerTests : ControllerTests
+    {
+        private DampAndMouldController _classUnderTest;
+        private Mock<IReportDampAndMouldUseCase> _reportDampAndMouldUseCase;
+
+        [SetUp]
```
LBHCallumM, line 17: I would also add a test for an invalid propertyReference

RepairsApi.Tests/E2ETests/DampAndMouldE2ETests.cs
```diff
@@ -0,0 +1,44 @@
...
+        private string _propertyReference;
+        private readonly Fixture _fixture = new Fixture();
+
+        [SetUp]
+        public void Setup()
+        {
+            _propertyReference = _fixture.Create<int>().ToString();
+            SetUserRole(UserGroups.Contractor);
+        }
+
```
LBHCallumM, line 24: Could we also add some tests for bad request scenarios

LBHCallumM (approved):
Make a few suggestions

RepairsApi/V2/Controllers/DampAndMouldController.cs
adamtry: I'm not sure either, but I was thinking that we're "adding reports to a property"

RepairsApi/V2/Controllers/DampAndMouldController.cs
adamtry: I simplified it instead, Regex was probably overkill https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1075/commits/c0a715911b32295de4c5aa21cc260271784a16dc

RepairsApi.Tests/V2/UseCase/ReportDampAndMouldUseCaseTests.cs
adamtry: The ReportedAt becomes slightly different because it takes a timestamp on creation. I moved the step to the separate tests file: https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1075/commits/92a59ea032571e8aeda93d8f9d386eb893b3db61

RepairsApi/V2/Gateways/DampAndMouldGatway.cs
adamtry: Good spot :smile:, fixed https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1075/commits/0c9ea8c5ef33e98c9111ba05ae229e8fa0f3b790

RepairsApi/V2/Gateways/DampAndMouldGatway.cs
adamtry: Good point - added in https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1075/commits/5d0d26b33ee39aa42140303fbc2f9f792d7c195c

RepairsApi.Tests/V2/Controllers/DampAndMouldControllerTests.cs
adamtry: Added here https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1075/commits/060f2fcc1883adbb0c2e1a754405ef84e480bb9a

RepairsApi.Tests/V2/Controllers/DampAndMouldControllerTests.cs
adamtry: True, I didn't. Removed in a later commit

RepairsApi.Tests/E2ETests/DampAndMouldE2ETests.cs
adamtry: I think those are somewhat redundant with the existing controller tests

PR #1076, opened 2024-01-11
Add GET damp and mould reports endpoint

adamtry:
[MR-869](https://hackney.atlassian.net/browse/MR-869) As a BSO, I need operatives to report damp and mould when completing a Repair

Adds a GET `/damp-and-mould/reports` query endpoint that supports queries with:
- propertyReference (default null)
- afterDate (default null)
- pageSize (default 30)
- page (default 1)

If a propertyReference is provided, returns any matching databse results for that property in reverse date order.

If no reference is provided, simply returns any database results in reverse date order.

If afterDate is provided, only returns reports that were created/raised on or after the provided UTC date, e.g. 2024-01-01

Pagination is applied to the database query as specified by the client query.

Other changes:
- Use java 17 jre for sonarcloud in the tests Dockerfile to allow SonarCloud to keep working - no longer supports java 11


RepairsApi/V2/Gateways/DampAndMouldGateway.cs
```diff
@@ -18,10 +22,29 @@ public DampAndMouldGateway(RepairsContext repairsContext, ILogger<DampAndMouldGa
...
+            DampAndMouldReportSearchParameters parameters)
+        {
+            _logger.LogInformation(
+                "Getting damp and mould reports page {Page} for {PropertyReference} propRef after date {AfterDate}",
+                parameters.PageNumber, parameters.PropertyReference ?? "any", parameters.AfterDate
+            );
+
+            var reports = _repairsContext.DampAndMouldReports
+                .Where(report => report.PropertyReference == parameters.PropertyReference || parameters.PropertyReference == null)
+                .Where(report => report.ReportedAt >= parameters.AfterDate || parameters.AfterDate == null)
```
LBHCallumM, line 42: How does this logic work? If null is passed for the property reference, does it ignore the condition?

RepairsApi.Tests/V2/Gateways/DampAndMouldGatewayTests.cs
```diff
@@ -52,5 +57,115 @@ public async Task CanCreateDampAndMouldReport()
             dbResponse?.Comments.Should().Be(report.Comments);
             dbResponse?.ReportedAt.Should().BeBefore(DateTime.UtcNow);
         }
+
+        [Test]
+        public async Task GetDampAndMouldReports_WhenPropRefProvided_ReturnsOnlyReportsForProperty()
+        {
+            // Arrange
+            var testReportCount = 10;
```
LBHCallumM, line 65: Can we add a test for when the propRef is null, to make sure any report is returned?

RepairsApi/V2/Gateways/DampAndMouldGateway.cs
LBHCallumM, line 42: Do we need to filter by the report date?

RepairsApi/V2/Gateways/DampAndMouldGateway.cs
```diff
@@ -18,10 +22,29 @@ public DampAndMouldGateway(RepairsContext repairsContext, ILogger<DampAndMouldGa
...
         {
-            _logger.LogInformation("Adding damp and mould report for property {ReportPropertyReference}", report.PropertyReference);
+            _logger.LogInformation("Adding damp and mould report for property {ReportPropertyReference}",
+                report.PropertyReference);
             await _repairsContext.DampAndMouldReports.AddAsync(report);
             await _repairsContext.SaveChangesAsync();
             return report.Id;
         }
+
+        public async Task<List<DampAndMouldReport>> GetDampAndMouldReports(
```
LBHCallumM, line 32: We will need to indicate to the frontend that there are more pages. 

For example, we could include something like this:
- pageNumber
- numberOfPages

RepairsApi.Tests/E2ETests/DampAndMouldE2ETests.cs
```diff
@@ -1,31 +1,50 @@
...
+        private IEnumerable<DampAndMouldReport> _reportsForProperty;
         private readonly Fixture _fixture = new Fixture();
 
-        [SetUp]
-        public void Setup()
+        [OneTimeSetUp]
+        public void OneTimeSetUp()
         {
-            _propertyReference = _fixture.Create<int>().ToString();
             SetUserRole(UserGroups.Contractor);
```
LBHCallumM, line 26: I think we should include a test to assert the authorization

RepairsApi/V2/Gateways/DampAndMouldGateway.cs
adamtry: Yes, it's ignored if null

adamtry: That's for the "Fetch records for last x days" functionality

LBHCallumM: I see. I imagined that logic being implemented by a subquery.

RepairsApi/V2/Gateways/DampAndMouldGateway.cs
adamtry: Makes sense, to allow numbering the pages on the FE.
Added in https://github.com/LBHackney-IT/repairs-api-dotnet/pull/1076/commits/55db916479d5b229418694318c3686a3676d93a1

RepairsApi.Tests/E2ETests/DampAndMouldE2ETests.cs
adamtry: Testing the authorisation of the endpoints seems unprecedented in this API, and it would mostly just be testing the same framework used in all the other endpoints - I don't think this is necessary

I did manually test this at least - the endpoints return unauthorized as expected without the correct group

LBHCallumM (approved):

PR #1080, opened 2024-01-16
Set and return address for damp & mould reports

adamtry:
[MR-869](https://hackney.atlassian.net/browse/MR-869)

We need to display the address of the damp and mould reports in the frontend - for this we'll need to store the address in the API when the report is raised and return it from the GET endpoint.

The address is provided by the client request object for raising a damp and mould report.


RepairsApi/V2/UseCase/ReportDampAndMouldUseCase.cs
```diff
@@ -1,23 +1,37 @@
...
 
         public async Task ExecuteAsync(DampAndMouldReportRequest request, string propertyReference)
         {
-            var report = request.ToDb(propertyReference);
+            var propertyResponse = await _assetGateway.GetByPropertyReference(propertyReference);
+            if (propertyResponse == null)
+            {
+                await _dampAndMouldGateway.AddDampAndMouldReport(request.ToDb(propertyReference, null));
+                return;
+            }
```
LBHCallumM, line 28: I think this needs to be a bad request scenario. It wont be the case that no property exists because a repair will have been raised against it.

adamtry: My concern is that Repairs Hub uses its own database for property references which doesn't necessarily always match the Asset API, and it's better to store a report without an address than to fail to store the report at all and prevent officers from completing their repair.

This would be an extremely rare case of course.

RepairsApi/V2/UseCase/ReportDampAndMouldUseCase.cs
```diff
@@ -1,23 +1,37 @@
...
+
+            var address = $"{propertyResponse.AssetAddress?.AddressLine1}, " +
+                               $"{propertyResponse.AssetAddress?.AddressLine2}, " +
+                               $"{propertyResponse.AssetAddress?.PostCode}";
```
LBHCallumM, line 32: Might this be a bit long? What if we used the same address used in the manage work orders page? That seems to work for users. I think that might just be addressline1.
image

LBHCallumM (approved):

